PR #60, opened 2026-04-27
feat: add EndpointInfo serializer and parser

### LINKED ISSUE #8: Implement EndpointInfo serializer + parser
## PROJECT CONTEXT

>
> **Reference material:**
>

---

### What and why
Quick Share peers identify themselves via a packed binary structure called **endpoint info**. The same byte layout appears in two places: as the value of the mDNS TXT record key `n` (URL-safe-base64 encoded) and as the `endpoint_info` field of the `ConnectionRequestFrame` proto (raw bytes).

### Layout (per PROTOCOL.md)
- 1 byte bitfield: `version(3 bits) | visibility(1 bit) | deviceType(3 bits) | reserved(1 bit)`.
  - Device types: 0=unknown, 1=phone, 2=tablet, 3=laptop, 4=car, 5=foldable, 6=XR.
  - Visibility: 0=visible, 1=hidden (in which case the device name is omitted).
- 16 bytes of "metadata" (2-byte salt + 14-byte encrypted-metadata-key). Random is fine — only meaningful if you've talked to Google's servers.
- 1 byte name length + UTF-8 device name (omitted if visibility=hidden).
- Optional TLV records: `1 byte type | 1 byte length | <length> bytes value`.
  - Type 1 = QR code data (advertising token, or AES-GCM ciphertext if hidden).
  - Type 2 = vendor ID (1 byte, 0=none, 1=Samsung).

### Acceptance criteria
- `EndpointInfo` data class with `serialize(): ByteArray` and `EndpointInfo.parse(bytes): EndpointInfo` (returns null on malformed).
- Round-trip property test (serialize → parse → equals original) over random valid inputs.
- Hidden-mode serialization omits the name field; parse handles that.
- Trailing TLV records preserved across round-trip.
- Reference implementation: `NearbyShare/NearbyConnectionManager.swift:EndpointInfo`.

**Parent:** #1

kyujin-cho:
## Summary

Implements the packed binary descriptor advertised by every Quick Share peer (`EndpointInfo`). Resolves #8.

The same byte layout appears in two places:
- mDNS TXT record key `n` (URL-safe-base64 encoded, no padding)
- `endpoint_info` field of `ConnectionRequestFrame` (raw bytes)

so bit-exact agreement with the NearDrop reference is required for interop.

## Wire format

```
byte 0    : version(3) | visibility(1) | deviceType(3) | reserved(1)   (MSB-first)
bytes 1-16: 2-byte salt + 14-byte encrypted-metadata-key
byte 17   : device-name length (omitted in hidden mode)
bytes 18+ : UTF-8 device name + optional TLV records (type|len|value)
```

## What changed

- `EndpointInfo` data class with `serialize()` and `EndpointInfo.parse(bytes)` returning `null` on malformed input (peers in the wild send junk, never throw).
- `DeviceType` enum mirroring spec values (`UNKNOWN`, `PHONE`, `TABLET`, `LAPTOP`, `CAR`, `FOLDABLE`, `XR`); `fromRaw` collapses unknown raw values (including the reserved `7`) to `UNKNOWN` so forward-compatible peers never get dropped.
- `TlvRecord` data class with structural equality on the value bytes, preserving unknown TLV types verbatim across round-trips.
- `Base64Url` helper (URL-safe alphabet, encode strips padding, decode tolerant of padding) for the mDNS TXT key `n` encoding.

## Test coverage

- Hand-crafted byte vectors that pin MSB-first bit packing of byte 0.
- Hidden-mode coverage (visibility bit set, name section absent).
- UTF-8 (Korean device name) round-trip; strict UTF-8 decoding rejects malformed sequences with `null`.
- Truncation cases (short header, missing name length, name overflow, TLV header truncation, TLV value overflow).
- Maximum-length name (255 bytes) and maximum-length TLV value (255 bytes).
- Unknown TLV types preserved across round-trip.
- Reserved deviceType raw value `7` collapses to `UNKNOWN`.
- Randomized round-trip property test (200 cases, fixed seed) over visibility, version, device type, reserved, metadata, mixed-script UTF-8 names, and random TLV records.
- Proto interop test that round-trips serialized bytes through `ConnectionRequestFrame.endpoint_info`.
- `Base64Url` round-trip including a serialized `EndpointInfo`.

## Test plan

- [x] `./gradlew :core-protocol:test`
- [x] `./gradlew build` (lint, ktlint, detekt all clean)

### DIFF
--- a/core-protocol/src/main/kotlin/io/github/kyujincho/wvmg/protocol/endpoint/Base64Url.kt
+++ b/core-protocol/src/main/kotlin/io/github/kyujincho/wvmg/protocol/endpoint/Base64Url.kt
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2026 WhenVivoMeetsGoogle contributors.
+ *
+ * Licensed under the Apache License, Version 2.0.
+ */
+package io.github.kyujincho.wvmg.protocol.endpoint
+
+import java.util.Base64
+
+/**
+ * URL-safe base64 codec used for the mDNS TXT record key `n`.
+ *
+ * Quick Share TXT records carry the [EndpointInfo] bytes encoded with the
+ * URL-safe alphabet (`-_` instead of `+/`) and **without** padding (`=`).
+ * The reference NearDrop encoder strips the trailing `=` characters before
+ * publishing the TXT record, so we mirror that behavior — both for round-trip
+ * symmetry and so peers that round-trip our advertisement back to us see the
+ * exact byte string they sent.
+ *
+ * The decoder accepts both padded and unpadded input (matching JDK
+ * `Base64.getUrlDecoder()`'s lenient stance) so we interoperate with peers
+ * that encode their advertisements with padding.
+ *
+ * Performance: the JDK's `Base64.Encoder` / `Base64.Decoder` are zero-cost
+ * thread-safe singletons; we cache them as `private val` to avoid the
+ * `getInstance` static lookup on every advertisement we publish.
+ */
+public object Base64Url {
+    private val ENCODER: Base64.Encoder = Base64.getUrlEncoder().withoutPadding()
+    private val DECODER: Base64.Decoder = Base64.getUrlDecoder()
+
+    /**
+     * Encodes [bytes] as URL-safe base64 **without** trailing `=` padding.
+     *
+     * The output is ASCII and safe to drop into a DNS TXT record value
+     * directly.
+     */
+    public fun encode(bytes: ByteArray): String = ENCODER.encodeToString(bytes)
+
+    /**
+     * Decodes a URL-safe base64 string. Padding is optional.
+     *
+     * Returns `null` on any decoding error (illegal characters, malformed
+     * length). Like [EndpointInfo.parse], this never throws on bad peer
+     * data — it just signals "ignore this advertisement".
+     */
+    public fun decode(input: String): ByteArray? =
+        try {
+            DECODER.decode(input)
+        } catch (_: IllegalArgumentException) {
+            null
+        }
+}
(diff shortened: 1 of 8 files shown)


kyujin-cho:

### Intent
Provide bit-exact `EndpointInfo` serializer + parser used in both mDNS TXT key `n` and `ConnectionRequestFrame.endpoint_info`, per PROTOCOL.md.

### Verification
- [x] All stated requirements implemented (data class, serialize/parse, hidden-mode, TLV preservation, randomized round-trip)
- [x] Implementation lands in `:core-protocol` per scope; downstream integration is the responsibility of #7 (mDNS) and the connection layer
- [x] Project conventions followed (license header, `explicitApi`, `public` qualifiers, KDoc style)
- [x] Existing modules reused (JDK `Base64`, `StandardCharsets`)
- [x] No unintended structural changes — new `endpoint` subpackage, no existing files touched
- [x] `./gradlew build` passes (Kotlin compile, ktlint, detekt, lint, all unit tests)

### Security & performance review
- Parser is fully bounds-checked: every offset advance is preceded by a length check, returning `null` instead of throwing on truncated/malformed input
- UTF-8 decode is strict (`CodingErrorAction.REPORT`) so malformed sequences fail closed rather than silently substituting U+FFFD
- No reflection, dynamic loading, or external I/O
- `Base64Url.decode` catches `IllegalArgumentException` and returns `null`
- Allocations are O(n) per parse/serialize; no quadratic loops

### Test coverage
46 test methods across `EndpointInfoTest`, `DeviceTypeTest`, `Base64UrlTest`, and `EndpointInfoProtoInteropTest`, including a 200-iteration randomized round-trip property test with mixed-script UTF-8 names and random TLV records.

Issue #24, opened 2026-04-27
Sender share intent handler (ACTION_SEND, ACTION_SEND_MULTIPLE)


kyujin-cho:
## PROJECT CONTEXT

> This project ports the **Google Nearby Share / Quick Share** protocol to Android, modeled on [NearDrop](https://github.com/grishka/NearDrop) — an open-source macOS implementation by @grishka. Quick Share is the over-the-air file-sharing protocol that ships with Android (delivered via Google Play Services) and is also implemented by Google's official Windows app and several third-party clients. Our goal is a standalone Android app that can send to and receive from any Quick Share peer **without depending on Google Play Services for the protocol logic**, making it usable on devices where GMS is absent or restricted.
>
> **Reference material:**
> - Protocol spec (read this first): https://github.com/grishka/NearDrop/blob/master/PROTOCOL.md
> - NearDrop source (the reference implementation we're porting from): https://github.com/grishka/NearDrop
> - Google's UKEY2 handshake spec: https://github.com/google/ukey2
> - Quick Share .proto files (vendored from Chromium): https://github.com/grishka/NearDrop/tree/master/NearbyShare/ProtobufSource
>
> **Scope:** Android-only port. Apple-side interop (AirDrop, AWDL, iPhone discovery) is explicitly **out of scope** for this project.

---

### What and why
On Android, the natural send-flow entry point is the system share sheet. When the user picks our app from "Share via", we receive an `ACTION_SEND` (single attachment) or `ACTION_SEND_MULTIPLE` intent, with a content URI (or list). Our sender activity then runs discovery, lets the user pick a target, and kicks off the outbound connection (#1.13).

### Acceptance criteria
- Manifest intent filters declaring all `ACTION_SEND`/`ACTION_SEND_MULTIPLE` MIME types our app supports (`*/*` is fine; restrict in code if needed).
- Activity reads `EXTRA_STREAM` (single Uri) or `EXTRA_STREAM` (ArrayList<Uri>) and `EXTRA_TEXT` (for plain-text shares).
- Calls `ContentResolver.takePersistableUriPermission` if the URI is persistable; otherwise resolves the file via temporary read access.
- Discovery runs while the picker is open: shows a list of nearby devices populated by mDNS browse (#1.14) and (Phase 2) BLE scan results.
- Selecting a device launches the outbound connection (#1.13).
- Handles plain-text share (single string in `EXTRA_TEXT`) as the URL/text payload type.

**Parent:** #1

PR #61, opened 2026-04-27
feat: add QR-code path generate, parse, and key derivation

### LINKED ISSUE #20: QR-code path: generate + parse, derive advertising/name keys
## PROJECT CONTEXT

>
> **Reference material:**
>

---

### What and why
The QR-code path is Quick Share's fallback discovery mechanism for cases where the receiver can't be made auto-visible (Samsung phones, OEMs that disabled BLE auto-trigger, etc.). The sender generates a QR code; the receiver scans it; the receiver then publishes its mDNS service with extra TLV data the sender recognises. This bypasses the BLE pulse entirely.

### URL format
`https://quickshare.google/qrcode#key=<urlsafe-base64-of-keyData>`

`keyData` = 3-byte version prefix `00 00 02` (or `03`) + 32-byte ECDSA P-256 X coordinate (compressed-form; sign byte handled separately).

### Key derivation (per PROTOCOL.md "QR codes" section)
```
advertisingToken    = HKDF-SHA256(ikm=keyData, salt="", info="advertisingContext", L=16)
nameEncryptionKey   = HKDF-SHA256(ikm=keyData, salt="", info="encryptionKey",      L=16)
```

### Receiver behavior when scanning QR
- Add a TLV record (type=1) to the EndpointInfo (#1.4). Contents depend on visibility:
  - **Visible**: TLV value = `advertisingToken` (16 bytes raw).
  - **Hidden**: visibility bit set; no plaintext name; TLV value = `IV(12) || AES-GCM ciphertext(name) || tag(16)`. Key = `nameEncryptionKey`. AAD = `advertisingToken`.

### Sender behavior when QR is active
- Compare each discovered device's TLV(type=1) against our `advertisingToken`. Match → connect immediately.
- If no match and TLV is >28 bytes, attempt AES-GCM decrypt with `nameEncryptionKey` and `advertisingToken` AAD; on success, the plaintext is the device name → connect.

### Acceptance criteria
- Sender side: generate ECDSA P-256 keypair, build URL, render to a QR bitmap (`com.google.zxing:core` or `io.github.g0dkar:qrcode-kotlin`).
- Receiver side: scan QR with CameraX + ZXing, parse URL, derive both keys, modify EndpointInfo with TLV before publishing.
- AES-GCM hidden-name path implemented and tested (encrypt on receiver, decrypt on sender).
- Be careful about EC X-coordinate leading-zero stripping (NearDrop calls this out: "Sometimes there will be a leading zero byte. Strip that, Android really hates it").

**Parent:** #1

kyujin-cho:
## Summary

Implements the pure-protocol pieces of Quick Share's QR-code fallback discovery path
(NearDrop `PROTOCOL.md` "QR codes" section). Closes #20.

A new package `:core-protocol/.../qr/` exposes everything the receiver UI (#24) and
sender connection logic will need:

- **`QrKeyData`** — 35-byte versioned payload (3-byte version prefix `00 00 02`/`03`
  + 32-byte ECDSA P-256 X coordinate). The `generate()` factory creates a fresh
  P-256 keypair and extracts the X coordinate as a fixed-width 32-byte unsigned
  integer, handling both `BigInteger.toByteArray()` edge cases (leading sign byte
  on high-bit-set values, short-coordinate left-padding) — NearDrop calls these
  out specifically: *"Sometimes there will be a leading zero byte. Strip that,
  Android really hates it."*
- **`QrUrl`** — build/parse for the canonical
  `https://quickshare.google/qrcode#key=<urlsafe-base64>` URL. The payload sits in
  the fragment (which never travels to a server), and the parser is strict: any
  deviation from the canonical scheme/host/path returns `null`.
- **`QrKeyDerivation`** — HKDF-SHA256 (atop the existing `Hkdf` from #9) to derive
  `advertisingToken` (`info="advertisingContext"`) and `nameEncryptionKey`
  (`info="encryptionKey"`), 16 bytes each.
- **`QrHiddenNameCipher`** — AES-128-GCM encrypt/decrypt with the on-the-wire
  layout `IV(12) || ciphertext || tag(16)` and `advertisingToken` bound as AAD.
  Decrypt returns `null` on every failure mode (auth tag mismatch, malformed
  UTF-8, short input) since the sender's matching loop iterates over many
  unrelated TLV records and exceptions per non-match would be noisy.
- **`QrTlvMatcher`** — sender-side helpers that scan a discovered
  `EndpointInfo`'s type=1 TLV records against our QR keys, supporting both
  visible-mode (16-byte raw advertising token compared via
  `MessageDigest.isEqual` for constant-time semantics) and hidden-mode
  (`> 28`-byte values fed through GCM decrypt to recover the device name).
  Also exposes receiver-side builders (`buildVisibleTlv`, `buildHiddenTlv`) so
  #24's scanner UI just composes them.

KAT vectors are pinned in `:core-protocol-test` (`QrCodeVectors`) so the same
expected outputs apply to both JVM unit tests and future Android instrumentation
tests; this avoids a class of bug where on-device tests silently re-derive the
"expected" outputs from the implementation under test.

QR-bitmap rendering (sender) and CameraX/ZXing scanning (receiver) are out of
scope for this issue per the spec — they land with the share intent in #24.

## Notes for reviewers

- The salt for both HKDF derivations is empty — that's mandated by the
  protocol, not a choice; per RFC 5869 §2.2 it expands internally to 32 zero
  bytes, which is what every Quick Share peer in the wild uses.
- The `> 28`-byte heuristic for hidden-mode decryption matches the issue body
  exactly: at exactly 28 bytes the ciphertext is empty (just IV + tag), which
  is not a useful match.
- `Cipher` is not thread-safe; every call allocates a fresh instance. Threads
  cannot share these helpers safely, but every call site is local-scope so
  this never matters in practice.

## Coverage

- `QrKeyDataTest` — wire layout, parse round-trip, EC X-coordinate edge cases
  (leading zero strip, left-pad, oversize reject), `generate()` returns a real
  P-256 keypair whose X coordinate matches the QR payload.
- `QrUrlTest` — round-trip, strict scheme/host/path/fragment validation,
  base64 alphabet sanity.
- `QrKeyDerivationTest` — three KAT vectors over the documented HKDF info
  strings, info-string identity, derived-key length, determinism, and that the
  two derived keys differ.
- `QrHiddenNameCipherTest` — encrypt/decrypt round-trip (ASCII, UTF-8, empty),
  IV freshness, AAD binding (flipping AAD breaks decrypt), tamper detection,
  short-input handling.
- `QrTlvMatcherTest` — visible match, hidden match, no-match, wrong-keys,
  unexpected-length TLV, multiple-TLV iteration, full receiver→sender flow
  for both visibility modes.


kyujin-cho:

### Intent
Implement the pure-protocol pieces of Quick Share's QR-code fallback discovery
(URL build/parse, key derivation, AES-GCM hidden-name cipher, sender-side
matching helpers) under `:core-protocol/.../qr/`.

### Verification
- [x] All stated requirements implemented (URL format, key data layout, both
      derived keys, AES-GCM round-trip, sender matching, receiver TLV builders)
- [x] Pure-protocol scope respected — QR-bitmap and CameraX work explicitly
      deferred to #24 per the issue body
- [x] Project conventions followed (Apache 2.0 headers, `explicitApi()`,
      Truth + JUnit5 testing, package layout mirrors existing
      `:core-protocol/.../endpoint/`)
- [x] Existing modules reused (`Hkdf` from #9, `Base64Url` and `EndpointInfo`
      / `TlvRecord` from #8) — no duplicated logic
- [x] No unintended structural changes (purely additive)
- [x] Tests pass (`./gradlew check` — full project green)

### Security & performance review
- [x] AES-GCM with 12-byte fresh-random IV (never reused), 128-bit auth tag
- [x] AAD (advertising token) bound to ciphertext — flipping it breaks decrypt
- [x] Constant-time comparison for advertising-token match
      (`MessageDigest.isEqual`)
- [x] Strict UTF-8 decoding (no U+FFFD substitution that could enable spoof)
- [x] No logging of sensitive material; KDoc explicitly warns
- [x] Defensive copy in `buildVisibleTlv` to prevent caller mutation
- [x] Single-pass `keyData.encode()` in `deriveKeys` to avoid redundant
      allocations
- [x] No CRITICAL or HIGH findings

### Test coverage
- 56 new tests across `QrKeyDataTest`, `QrUrlTest`, `QrKeyDerivationTest`,
  `QrHiddenNameCipherTest`, `QrTlvMatcherTest`
- 3 KAT vectors pinned in `:core-protocol-test` (`QrCodeVectors`) for both
  HKDF info strings and derived-key bytes

Issue #11, opened 2026-04-27
Derive D2D + SecureMessage encryption/HMAC keys from UKEY2 nextSecret


kyujin-cho:
## PROJECT CONTEXT

> This project ports the **Google Nearby Share / Quick Share** protocol to Android, modeled on [NearDrop](https://github.com/grishka/NearDrop) — an open-source macOS implementation by @grishka. Quick Share is the over-the-air file-sharing protocol that ships with Android (delivered via Google Play Services) and is also implemented by Google's official Windows app and several third-party clients. Our goal is a standalone Android app that can send to and receive from any Quick Share peer **without depending on Google Play Services for the protocol logic**, making it usable on devices where GMS is absent or restricted.
>
> **Reference material:**
> - Protocol spec (read this first): https://github.com/grishka/NearDrop/blob/master/PROTOCOL.md
> - NearDrop source (the reference implementation we're porting from): https://github.com/grishka/NearDrop
> - Google's UKEY2 handshake spec: https://github.com/google/ukey2
> - Quick Share .proto files (vendored from Chromium): https://github.com/grishka/NearDrop/tree/master/NearbyShare/ProtobufSource
>
> **Scope:** Android-only port. Apple-side interop (AirDrop, AWDL, iPhone discovery) is explicitly **out of scope** for this project.

---

### What and why
After UKEY2 finalizes, both sides compute the same shared secret (`SHA256(ECDH(peer_pub, our_priv).x_magnitude)`) and from it derive **two HKDF outputs**: `authString` (used for the PIN) and `nextSecret` (used for traffic keys). From `nextSecret`, four 32-byte AES-256 keys must be derived, two per direction.

### Derivation chain (per `NearbyConnection.swift:finalizeKeyExchange` and PROTOCOL.md)
```
ukeyInfo = ukeyClientInitMsg || ukeyServerInitMsg          (raw serialized Ukey2Message bytes; NOT length-prefixed)

authString    = HKDF-SHA256(ikm=dhs, salt="UKEY2 v1 auth", info=ukeyInfo, L=32)
nextSecret    = HKDF-SHA256(ikm=dhs, salt="UKEY2 v1 next", info=ukeyInfo, L=32)

D2D_salt      = SHA256("D2D")  =  0x82AA55A0D397F88346CA1CEE8D3909B95F13FA7DEB1D4AB38376B8256DA85510
[...]

PR #64, opened 2026-04-27
feat: derive 4-digit PIN from UKEY2 authString (Chromium algorithm)

### LINKED ISSUE #12: Derive 4-digit PIN from UKEY2 authString (Chromium algorithm)
## PROJECT CONTEXT

>
> **Reference material:**
>

---

### What and why
Quick Share displays a **4-digit PIN** on both peers so the user can visually confirm there's no MITM. The algorithm is **not** the one specified by UKEY2 — it's a custom Chromium one. Both peers must produce the same digits for the same authString.

### Algorithm (per `NearbyConnection.swift:pinCodeFromAuthKey`)
```
hash   = 0
mult   = 1
for byte b in authString.bytes (interpreted as SIGNED int8):
    hash = (hash + b * mult) mod 9973
    mult = (mult * 31) mod 9973
return "%04d".format(abs(hash))
```

### Java/Kotlin gotcha
`byte` is signed in Java already, but `ByteArray.get(i)` returns `Byte` which auto-converts to `Int` with sign extension when used in arithmetic. So `(byte.toInt())` is the correct conversion. Don't forget `Math.abs()` and zero-pad to 4 digits.

### Acceptance criteria
- `pinFromAuthString(authString: ByteArray): String` returns exactly 4 digits.
- KAT: a known authString → known PIN. Cross-check by running NearDrop with a hardcoded authString.

**Parent:** #1

kyujin-cho:
## Summary

- Adds `PinDerivation.deriveFourDigitPin(authString: ByteArray): String` in `:core-protocol/.../crypto/pin/`, a pure-Kotlin port of NearDrop's `NearbyConnection.swift:pinCodeFromAuthKey`. Both Quick Share peers must produce the same digits for the same `authString`, so the implementation is bit-exact with the Swift reference.
- Adds 10 Known-Answer Test vectors in `:core-protocol-test` (`PinVectors`) so the same answer-table can be reused from Android instrumentation tests later in Phase 1.

## Why

Quick Share displays a 4-digit PIN on both peers after the UKEY2 handshake so the user can visually confirm there is no MITM. The algorithm is a Chromium-specific construction that is **not** the one specified by UKEY2 itself; deriving the PIN incorrectly is silent — both peers would proceed with the connection and the user could approve a transfer that another device is waiting on. The KAT table guards against the two failure modes that would not show up in a single-peer smoke test:

- **Sign extension.** Swift uses `Int(Int8(bitPattern: _byte))`. The JVM's `Byte.toInt()` already sign-extends, so `b.toInt()` is sufficient — but only if the conversion goes through `Byte.toInt()` and not `b.toInt() and 0xFF`, which would zero-extend. Vectors with `0xFF` bytes lock this down.
- **Truncated-remainder modulo.** Swift's `%` and Kotlin's `%` on `Int` both follow the sign of the dividend. Languages with floored division (Python's `%`, Lua's `%`) would diverge once `hash` goes negative. Multi-byte vectors that wrap the modulus several times catch any drift here.

## Algorithm

Reproduced from `NearbyConnection.swift:pinCodeFromAuthKey` for clarity:

```
hash       = 0
multiplier = 1
for each byte b (interpreted as signed int8):
    hash       = (hash + b * multiplier) mod 9973
    multiplier = (multiplier * 31) mod 9973
return "%04d" % abs(hash)
```

`9973` is the largest prime less than `10000`, so `abs(hash) < 10000` always fits in 4 decimal digits.

## Public API

- `io.github.kyujincho.wvmg.protocol.crypto.pin.PinDerivation.deriveFourDigitPin(authString: ByteArray): String`
- Returns exactly 4 ASCII digits (left-padded with `'0'`).
- Empty input returns `"0000"` — matches the Swift reference and removes the need for callers to pre-validate.
- Pure function: no logging, no mutation of the input buffer, deterministic.

## KAT Vectors

`:core-protocol-test/.../PinVectors.kt` exposes 10 vectors covering:

| Vector | Purpose |
|---|---|
| empty authString | loop-boundary + zero-padding |
| single 0x00 byte | one-iteration baseline |
| single 0x01 byte | trivial positive path |
| single 0xFF byte | sign-extension guard |
| two 0xFF bytes | sign extension across iterations |
| 0x01 0x02 | multiplier-step verification |
| 32x 0x7F | multi-wrap, all positive |
| 32x 0xFF | multi-wrap, all negative |
| 0x00..0x1F | strictly increasing positives |
| 0x80,0x7F repeating | alternating extremes, hash swings both signs |

All values were computed from a faithful reproduction of the Swift algorithm and cross-checked against the NearDrop source.

## Notes

- `:core-protocol` stays free of `android.*` dependencies — uses only `kotlin.math.abs` from the stdlib.
- Issue #11 (running in parallel) will surface the actual `authString` from the UKEY2 handshake; this PR does not depend on its API. Callers will pass the bytes returned by `Ukey2HandshakeResult` once both PRs land.

Closes #12.

### DIFF
--- a/core-protocol-test/src/main/kotlin/io/github/kyujincho/wvmg/protocol/test/PinVectors.kt
+++ b/core-protocol-test/src/main/kotlin/io/github/kyujincho/wvmg/protocol/test/PinVectors.kt
@@ -1,0 +1,203 @@
+/*
+ * Copyright 2026 WhenVivoMeetsGoogle contributors.
+ *
+ * Licensed under the Apache License, Version 2.0.
+ */
+package io.github.kyujincho.wvmg.protocol.test
+
+/**
+ * Known-Answer Test (KAT) vectors for the Quick Share 4-digit PIN derivation
+ * defined in `NearbyConnection.swift:pinCodeFromAuthKey`.
+ *
+ * These vectors are computed from a faithful reproduction of the Swift
+ * algorithm and cross-checked against the NearDrop reference source at
+ * https://github.com/grishka/NearDrop/blob/master/NearbyShare/NearbyConnection.swift.
+ *
+ * The vectors are grouped to lock down three things at once:
+ *
+ *  1. **Boundary cases** — empty input, all-zero input, the single-byte
+ *     baseline. These keep the loop boundary and `"%04d"` zero-padding
+ *     honest.
+ *  2. **Sign extension** — at least one vector contains a byte with the
+ *     high bit set (`0xFF`). The Swift reference uses
+ *     `Int(Int8(bitPattern: _byte))`, which produces `-1` for `0xFF`. The
+ *     equivalent on the JVM is `Byte.toInt()` (already sign-extending). A
+ *     buggy port that uses `b.toInt() and 0xFF` would produce a different
+ *     PIN here, and this vector is what catches that regression.
+ *  3. **Modulo wrap-around** — vectors long enough (32 bytes) that the
+ *     intermediate `hash` and `multiplier` wrap past the 9973 modulus
+ *     several times, exercising the truncated-remainder contract that
+ *     differs between Swift/Kotlin/Java (sign-of-dividend) and
+ *     Python/Lua (floored).
+ *
+ * The KAT vectors live in `:core-protocol-test` so the same answer-table can
+ * be reused from Android instrumentation tests when the protocol stack
+ * lights up later in Phase 1.
+ */
+public object PinVectors {
+    /**
+     * One PIN derivation KAT vector.
+     *
+     * Intentionally a regular `class` rather than a `data class`: the
+     * auto-generated `equals`/`hashCode` for `ByteArray` use reference
+     * identity (a known JVM gotcha) so the data-class default is broken,
+     * and we do not need destructuring or `copy()` for read-only fixtures.
+     * `toString` is overridden to return the human-readable [name] so test
+     * failures stay legible.
+     */
+    public class Vector(
+        public val name: String,
+        public val authString: ByteArray,
+        public val expectedPin: String,
+    ) {
+        init {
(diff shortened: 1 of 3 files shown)


kyujin-cho:

### Intent
> Port NearDrop's `pinCodeFromAuthKey` (Chromium's 4-digit PIN derivation) to pure Kotlin so both Quick Share peers can compute the same visual confirmation digits from the UKEY2 `authString`.

### Findings Addressed
_None — no CRITICAL or HIGH severity findings._

### Verification
- [x] All stated requirements implemented (signed-byte interpretation, truncated mod, abs, 4-digit zero-pad)
- [x] Algorithm is bit-exact with NearDrop's `NearbyConnection.swift:pinCodeFromAuthKey` (verbatim quoted in the implementation docstring; cross-checked against the reference source)
- [x] Project conventions followed (file header, `public object`, `explicitApi`, KAT vectors split into `:core-protocol-test` matching `HkdfVectors`)
- [x] Existing modules reused where applicable (pure stdlib `kotlin.math.abs` — nothing to reuse for an isolated arithmetic function)
- [x] No unintended structural changes — only adds files
- [x] Tests pass: 15 `PinDerivationTest` cases (10 KAT vectors + 5 invariant tests)
- [x] `staticAnalysis` (ktlint + detekt) clean

### Security & Performance
- No untrusted input parsing; `authString` is post-handshake opaque bytes
- O(n) with constant-time per-byte ops; bounded `Int` arithmetic (`abs(intermediate) < 9973 * 127 < 2^21`)
- No logging of `authString` or intermediate state by design
- Pure deterministic function; no I/O, no allocations beyond the output `String`
- Input buffer is not mutated (verified by test)

Integration into the connection flow lands in a follow-up PR once #11 exposes `Ukey2HandshakeResult.authString`.

Issue #28, opened 2026-04-27
Loopback integration test (sender + receiver in-process)


kyujin-cho:
## PROJECT CONTEXT

> This project ports the **Google Nearby Share / Quick Share** protocol to Android, modeled on [NearDrop](https://github.com/grishka/NearDrop) — an open-source macOS implementation by @grishka. Quick Share is the over-the-air file-sharing protocol that ships with Android (delivered via Google Play Services) and is also implemented by Google's official Windows app and several third-party clients. Our goal is a standalone Android app that can send to and receive from any Quick Share peer **without depending on Google Play Services for the protocol logic**, making it usable on devices where GMS is absent or restricted.
>
> **Reference material:**
> - Protocol spec (read this first): https://github.com/grishka/NearDrop/blob/master/PROTOCOL.md
> - NearDrop source (the reference implementation we're porting from): https://github.com/grishka/NearDrop
> - Google's UKEY2 handshake spec: https://github.com/google/ukey2
> - Quick Share .proto files (vendored from Chromium): https://github.com/grishka/NearDrop/tree/master/NearbyShare/ProtobufSource
>
> **Scope:** Android-only port. Apple-side interop (AirDrop, AWDL, iPhone discovery) is explicitly **out of scope** for this project.

---

### What and why
Before testing against external Quick Share peers, prove our two state machines speak to each other. Run sender and receiver in the same JVM/process, connected via in-memory pipes (or `localhost` sockets), and exercise the full file-transfer flow.

### Acceptance criteria
- A test that spawns a `ReceiverServer` on a random port and an `OutboundConnection` connecting to it.
- Transfers a 1 KB file, a 1 MB file, and a 100 MB file (the last is performance-sensitive — should complete in under 30 s on a modern dev machine).
- Verifies SHA-256 of received bytes matches sender.
- Tests cancel mid-transfer from sender.
- Tests cancel mid-transfer from receiver.
- Tests reject (receiver declines).
- Tests text/URL transfer.

**Parent:** #1

Issue #21, opened 2026-04-27
Foreground receiver service (connectedDevice type)


kyujin-cho:
## PROJECT CONTEXT

> This project ports the **Google Nearby Share / Quick Share** protocol to Android, modeled on [NearDrop](https://github.com/grishka/NearDrop) — an open-source macOS implementation by @grishka. Quick Share is the over-the-air file-sharing protocol that ships with Android (delivered via Google Play Services) and is also implemented by Google's official Windows app and several third-party clients. Our goal is a standalone Android app that can send to and receive from any Quick Share peer **without depending on Google Play Services for the protocol logic**, making it usable on devices where GMS is absent or restricted.
>
> **Reference material:**
> - Protocol spec (read this first): https://github.com/grishka/NearDrop/blob/master/PROTOCOL.md
> - NearDrop source (the reference implementation we're porting from): https://github.com/grishka/NearDrop
> - Google's UKEY2 handshake spec: https://github.com/google/ukey2
> - Quick Share .proto files (vendored from Chromium): https://github.com/grishka/NearDrop/tree/master/NearbyShare/ProtobufSource
>
> **Scope:** Android-only port. Apple-side interop (AirDrop, AWDL, iPhone discovery) is explicitly **out of scope** for this project.

---

### What and why
For the receiver to discover senders and accept transfers while the user has navigated away from the app or turned the screen off, we need an Android **foreground service**. Android 14+ requires explicit foreground-service types — `connectedDevice` is the right one for our use case (peer-to-peer file transfer over Wi-Fi/Bluetooth).

### Acceptance criteria
- Service class extending `Service` with `START_STICKY` semantics.
- `foregroundServiceType="connectedDevice"` declared in manifest (API 34+).
- Persistent notification with channel `low_importance` showing "Quick Share is active".
- Service starts the receiver server (#1.15) + mDNS advertise (#1.14) on `onStartCommand`.
- Stops cleanly on `onDestroy`, releases all resources (multicast lock, sockets, JmDNS, BLE in Phase 2).
- Survives screen-off, low-memory pressure (foreground services are last-killed).

**Parent:** #1

Issue #18, opened 2026-04-27
mDNS publish + browse via JmDNS


kyujin-cho:
## PROJECT CONTEXT

> This project ports the **Google Nearby Share / Quick Share** protocol to Android, modeled on [NearDrop](https://github.com/grishka/NearDrop) — an open-source macOS implementation by @grishka. Quick Share is the over-the-air file-sharing protocol that ships with Android (delivered via Google Play Services) and is also implemented by Google's official Windows app and several third-party clients. Our goal is a standalone Android app that can send to and receive from any Quick Share peer **without depending on Google Play Services for the protocol logic**, making it usable on devices where GMS is absent or restricted.
>
> **Reference material:**
> - Protocol spec (read this first): https://github.com/grishka/NearDrop/blob/master/PROTOCOL.md
> - NearDrop source (the reference implementation we're porting from): https://github.com/grishka/NearDrop
> - Google's UKEY2 handshake spec: https://github.com/google/ukey2
> - Quick Share .proto files (vendored from Chromium): https://github.com/grishka/NearDrop/tree/master/NearbyShare/ProtobufSource
>
> **Scope:** Android-only port. Apple-side interop (AirDrop, AWDL, iPhone discovery) is explicitly **out of scope** for this project.

---

### What and why
Quick Share peers find each other on Wi-Fi LAN via **multicast DNS** (Bonjour). The receiver advertises a service of type `_FC9F5ED42C8A._tcp.` on an arbitrary TCP port, with a specific instance name and a single TXT record key `n` whose value is the URL-safe-base64-encoded EndpointInfo (#1.4).

Android's built-in `NsdManager` has historically had bugs with TXT-record support, so we'll use **JmDNS** (`org.jmdns:jmdns`), the long-standing Java implementation, for both publishing and browsing. It works identically on all API levels and gives full control over TXT records.

### Service-instance name format
The mDNS instance name must be 10 bytes URL-safe-base64-encoded:
- `0x23` (PCP)
- 4 random alphanumeric ASCII bytes (endpoint ID)
- `0xFC, 0x9F, 0x5E` (the first 3 bytes of `SHA256("NearbySharing")` — the service ID hash)
- `0x00, 0x00`

### Acceptance criteria
- Wraps JmDNS publish in a coroutine-friendly API (`Discovery.advertise(endpointInfo, port): AdvertiseHandle`).
- Wraps JmDNS browse similarly (`Discovery.browse(): Flow<DiscoveredService>`).
[...]

Issue #16, opened 2026-04-27
Inbound (receiver) connection state machine


kyujin-cho:
## PROJECT CONTEXT

> This project ports the **Google Nearby Share / Quick Share** protocol to Android, modeled on [NearDrop](https://github.com/grishka/NearDrop) — an open-source macOS implementation by @grishka. Quick Share is the over-the-air file-sharing protocol that ships with Android (delivered via Google Play Services) and is also implemented by Google's official Windows app and several third-party clients. Our goal is a standalone Android app that can send to and receive from any Quick Share peer **without depending on Google Play Services for the protocol logic**, making it usable on devices where GMS is absent or restricted.
>
> **Reference material:**
> - Protocol spec (read this first): https://github.com/grishka/NearDrop/blob/master/PROTOCOL.md
> - NearDrop source (the reference implementation we're porting from): https://github.com/grishka/NearDrop
> - Google's UKEY2 handshake spec: https://github.com/google/ukey2
> - Quick Share .proto files (vendored from Chromium): https://github.com/grishka/NearDrop/tree/master/NearbyShare/ProtobufSource
>
> **Scope:** Android-only port. Apple-side interop (AirDrop, AWDL, iPhone discovery) is explicitly **out of scope** for this project.

---

### What and why
The receiver-side glue tying together framing, UKEY2, SecureMessage, and the negotiation state machine. Wraps a single `Socket` and represents one in-flight transfer attempt.

Reference: `NearbyShare/InboundNearbyConnection.swift` (the entire file).

### Behavior
- Accept incoming TCP connection.
- Drive the receiver-role state machine through to `waitingForUserConsent`.
- Surface the `TransferMetadata` (filenames, sizes, MIME types, PIN) to UI.
- Wait for `submitUserConsent(accepted: Bool)` from UI.
- On accept: send `ConnectionResponseFrame{status=ACCEPT}`, open files in Downloads (using `MediaStore` — see #1.21), stream chunks, send `Disconnection` and close on completion.
- On reject: send `ConnectionResponseFrame{status=REJECT}`, send `Disconnection`, close.

### Acceptance criteria
- Class `InboundConnection(socket: Socket)` with delegate-style callbacks: `onConsentNeeded(metadata)`, `onProgress(percent)`, `onCompleted`, `onFailed(reason)`.
- Submit-consent method (`submitUserConsent`) is thread-safe.
- Cancel from peer mid-transfer triggers partial-file cleanup (delete).
[...]

PR #75, opened 2026-04-27
feat: add TCP receiver server with per-connection coroutine lifecycle

### LINKED ISSUE #19: TCP server socket + per-connection coroutine lifecycle
## PROJECT CONTEXT

>
> **Reference material:**
>

---

### What and why
The receiver listens on an arbitrary TCP port (announced via mDNS). Each accepted connection runs through the inbound state machine (#1.12). We use coroutines for I/O — one supervisor scope per connection — so cancellation propagates cleanly when the user dismisses a transfer or the service stops.

### Acceptance criteria
- `ReceiverServer` class that opens a `ServerSocket` on port 0 (kernel picks), exposes the chosen port to the discovery layer.
- Each accepted `Socket` runs in `supervisorScope { ... }` on `Dispatchers.IO`.
- TCP_NODELAY enabled on every socket (matches NearDrop's `tcp.noDelay = true` in `startOutgoingTransfer`).
- Stopping the service cancels all in-flight connections and closes the listener.
- Per-connection mutex ensures no concurrent sends (sequence-number invariant, see #1.9).

**Parent:** #1

kyujin-cho:
## Summary

Closes #19.

Adds `TcpReceiverServer` in `:core-protocol` that owns a kernel-assigned ephemeral TCP port, runs an `accept()` loop on `Dispatchers.IO`, and hands each accepted `Socket` to a fresh `InboundConnection` (#16). The chosen port is what mDNS (#18) advertises in its Quick Share SRV record.

### Design

- **Module choice**: `:core-protocol`. The implementation is pure-JVM (no `android.*` imports) — it uses only `java.net.ServerSocket` / `Socket` / coroutine I/O — so it lives next to the rest of the protocol stack. Android-side wiring (foreground service, MediaStore-backed `FileDestinationFactory`) wraps this from `:service-android` (#21).
- **Concurrency**: A `SupervisorJob` child of the caller's scope owns the accept loop and every per-connection coroutine. `SupervisorJob` (not a plain `Job`) ensures one failing transfer cannot cancel siblings — Quick Share permits parallel inbound connections from different peers.
- **Cancellation**: `accept()` is wrapped in `runInterruptible(Dispatchers.IO)` so coroutine cancellation eagerly interrupts the blocked thread. Closing the `ServerSocket` from `stop()` additionally surfaces a `SocketException` which the loop treats as the graceful exit signal. `stop()` also eagerly closes every live client socket because `InboundConnection`'s blocking reads under `withContext(Dispatchers.IO)` do not honour coroutine cancellation while parked in a syscall — closing the socket is the only reliable way to unblock.
- **TCP_NODELAY**: Enabled on every accepted socket (matches NearDrop's `tcp.noDelay = true`) so small protocol frames are not Nagle-delayed.
- **Per-connection mutex**: Defense-in-depth against accidental concurrent `InboundConnection.run` invocation, upholding the sequence-number invariant.

### Public API

- `start()` — binds an ephemeral port, spawns the accept loop, returns the bound port.
- `stop()` / `stopBlocking()` — closes the listener, eagerly closes live client sockets, cancels every per-connection coroutine, joins.
- `boundPort` — the chosen port; valid only between `start` and `stop`.
- `results: SharedFlow<InboundConnectionCompletion>` — one terminal outcome per accepted connection.
- `activeConnections: SharedFlow<InboundConnection>` — newly-accepted connections; UIs subscribe to attach to each connection's `state` flow.

### Tests

Unit tests in `TcpReceiverServerTest` use real `ServerSocket` on the loopback. They cover port binding, reachability, idempotent `stop`, single-shot `start`, and `boundPort` lifecycle. Tests that pair the accept loop with the production `InboundConnection` are `@Disabled` with a pointer to #28 (loopback inbound + outbound integration).

## Test plan

- [x] `./gradlew :core-protocol:test` passes (4 tests run, 5 deferred to #28)
- [x] `./gradlew :core-protocol:ktlintCheck :core-protocol:detekt` passes
- [x] No `android.*` imports added to `:core-protocol`

### DIFF
--- a/core-protocol/src/main/kotlin/io/github/kyujincho/wvmg/protocol/server/TcpReceiverServer.kt
+++ b/core-protocol/src/main/kotlin/io/github/kyujincho/wvmg/protocol/server/TcpReceiverServer.kt
@@ -1,0 +1,458 @@
+/*
+ * Copyright 2026 WhenVivoMeetsGoogle contributors.
+ *
+ * Licensed under the Apache License, Version 2.0.
+ */
+package io.github.kyujincho.wvmg.protocol.server
+
+import io.github.kyujincho.wvmg.protocol.connection.InboundConnection
+import io.github.kyujincho.wvmg.protocol.connection.InboundResult
+import io.github.kyujincho.wvmg.protocol.payload.FileDestinationFactory
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancelAndJoin
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.flow.asSharedFlow
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.runInterruptible
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlinx.coroutines.withContext
+import java.io.IOException
+import java.net.InetAddress
+import java.net.ServerSocket
+import java.net.Socket
+import java.net.SocketException
+import java.security.SecureRandom
+import java.util.concurrent.atomic.AtomicLong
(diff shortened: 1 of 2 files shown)


kyujin-cho:
Duplicate of #74 (already merged via the orchestrator's manual recovery while this agent was still running). Closing to avoid double-applying.

Issue #34, opened 2026-04-27
mDNS gating: only publish while sender pulse is active


kyujin-cho:
## PROJECT CONTEXT

> This project ports the **Google Nearby Share / Quick Share** protocol to Android, modeled on [NearDrop](https://github.com/grishka/NearDrop) — an open-source macOS implementation by @grishka. Quick Share is the over-the-air file-sharing protocol that ships with Android (delivered via Google Play Services) and is also implemented by Google's official Windows app and several third-party clients. Our goal is a standalone Android app that can send to and receive from any Quick Share peer **without depending on Google Play Services for the protocol logic**, making it usable on devices where GMS is absent or restricted.
>
> **Reference material:**
> - Protocol spec (read this first): https://github.com/grishka/NearDrop/blob/master/PROTOCOL.md
> - NearDrop source (the reference implementation we're porting from): https://github.com/grishka/NearDrop
> - Google's UKEY2 handshake spec: https://github.com/google/ukey2
> - Quick Share .proto files (vendored from Chromium): https://github.com/grishka/NearDrop/tree/master/NearbyShare/ProtobufSource
>
> **Scope:** Android-only port. Apple-side interop (AirDrop, AWDL, iPhone discovery) is explicitly **out of scope** for this project.

---

### What and why
Stock Quick Share doesn't broadcast mDNS continuously — only while a sender's BLE pulse is detected. We match this behavior to (1) save Wi-Fi/CPU and (2) avoid interfering with users who don't want to be visible. Refactor #1.14 (mDNS publish) so it's driven by the BLE scanner state from #2.3.

### Acceptance criteria
- Receiver service maintains a `mdnsActive` flag toggled by BLE scan callbacks.
- mDNS publish call goes out only when `mdnsActive=true`.
- After mDNS goes up, it stays up for at least 30 s (debounce) to avoid flapping if the BLE pulse is intermittent.
- Manual override: user can force-enable visibility from the app's UI (so this still works on devices where BLE scanning is disabled).
- QR-code path overrides this gating entirely — see #1.16.

**Parent:** #2

Issue #33, opened 2026-04-27
Receiver BLE scanner with ScanFilter


kyujin-cho:
## PROJECT CONTEXT

> This project ports the **Google Nearby Share / Quick Share** protocol to Android, modeled on [NearDrop](https://github.com/grishka/NearDrop) — an open-source macOS implementation by @grishka. Quick Share is the over-the-air file-sharing protocol that ships with Android (delivered via Google Play Services) and is also implemented by Google's official Windows app and several third-party clients. Our goal is a standalone Android app that can send to and receive from any Quick Share peer **without depending on Google Play Services for the protocol logic**, making it usable on devices where GMS is absent or restricted.
>
> **Reference material:**
> - Protocol spec (read this first): https://github.com/grishka/NearDrop/blob/master/PROTOCOL.md
> - NearDrop source (the reference implementation we're porting from): https://github.com/grishka/NearDrop
> - Google's UKEY2 handshake spec: https://github.com/google/ukey2
> - Quick Share .proto files (vendored from Chromium): https://github.com/grishka/NearDrop/tree/master/NearbyShare/ProtobufSource
>
> **Scope:** Android-only port. Apple-side interop (AirDrop, AWDL, iPhone discovery) is explicitly **out of scope** for this project.

---

### What and why
The receiver listens for the sender's BLE pulse from #2.2 to know when to wake up its own mDNS service. Without seeing a pulse, the receiver stays silent — saving battery and matching stock Quick Share behavior.

### Implementation
```kotlin
val scanner = bluetoothAdapter.bluetoothLeScanner
val filter = ScanFilter.Builder()
    .setServiceData(
        ParcelUuid.fromString("0000fe2c-0000-1000-8000-00805f9b34fb"),
        byteArrayOf(0xfc, 0x12, 0x8e, 0x01, 0x42, ...),  // prefix only
        byteArrayOf(0xff, 0xff, 0xff, 0xff, 0xff, ...)   // mask: match prefix bytes only
    )
    .build()
val settings = ScanSettings.Builder()
    .setScanMode(ScanSettings.SCAN_MODE_BALANCED)
    .setReportDelay(0)
    .build()
scanner.startScan(listOf(filter), settings, callback)
```

### Acceptance criteria
- Scanner starts on receiver service start.
- On each matching scan result, gates the mDNS publish (#2.4).
- Scanner stops after a configurable inactivity timeout (e.g. 30 s after last seen pulse).
[...]

PR #97, opened 2026-04-28
fix: Phase 2 real-device interop bugs found via Vivo X300 Ultra testing

kyujin-cho:
## Summary

On-device verification of Phase 2 (BLE auto-discovery) against a Vivo X300 Ultra — Funtouch 16 / OriginOS 6 / Android 16 — surfaced four concrete issues that prevented the BLE pulse and mDNS pipeline from actually working end-to-end. Three are real production bugs that affect every device; the fourth is a Funtouch-specific developer-experience improvement.

PR #91 / #92 / #93 / #94 / #95 / #96 (Phase 2 itself) had landed CI-green and JVM-test-clean, but **CI has been failing for billing reasons** (`recent account payments have failed or your spending limit needs to be increased`) on every recent run, and the JVM unit tests for the BLE advertiser asserted on payload bytes individually rather than round-tripping through the platform's 31-byte budget. So none of these regressions had a unit-test or CI signal — they only became visible on real hardware.

### 1. `DATA_TOO_LARGE` — sender BLE pulse never broadcasts (607c41e)

`BleAdvertiser.buildAdvertiseData` called both `addServiceUuid(parcelUuid)` and `addServiceData(parcelUuid, payload)`. The 16-bit Quick Share UUID is already embedded inside the service-data AD (type `0x16`), and the receiver-side `ScanFilter` from #33 matches on service data only — the separate UUID-list AD had no consumer. Including both produced a 32-byte advertising payload (4 + 28), one byte over the 31-byte legacy BLE budget. The platform reported `onStartFailure code=1 (DATA_TOO_LARGE)` and never broadcast the pulse, so stock Quick Share receivers never saw the trigger.

After dropping the redundant `addServiceUuid`, the dump shows the advertise running cleanly:

```
io.github.kyujincho.wvmg.debug
  Advertising: Legacy=true Connectable=false TX_POWER=1 dBm
  Service Data(UUID, length of data) : [0000fe2c-..., 24]
  Elapsed time: 167573ms (no failures)
```

### 2. Receiver foreground service was never started (81863dd)

`ReceiverForegroundService.start(context)` had no production caller. The companion entrypoint existed but only tests invoked it. Consequently, on a freshly installed device the BLE pulse scanner (#33), mDNS-publish gate (#34), and TCP listener were running only inside instrumentation tests; opening the launcher and waiting did not actually make the device discoverable. PR #89's Galaxy S26 Ultra interop work exercised the *send* path end-to-end, which is why this gap on the receive side went unnoticed until the BLE-trigger runbook was actually exercised on a real device.

`MainActivity.onStart` is the right place: the user is visibly in our app (so Android 14+ allows starting a foreground service), the permission gate has either passed or routed the user to onboarding, and `startService` is idempotent so repeated `onStart`s cost nothing.

### 3. mDNS multicast lock silently refused on Funtouch (e57b0ae)

Vivo / Funtouch / OriginOS routinely treats `WifiManager.MulticastLock.acquire` as a no-op for non-system apps. Symptom: the initial JmDNS announce burst still goes out (UDP send is allowed) so peers see us briefly, but ongoing PTR queries from peers receive no response, and peers cache us for ~30 s before we vanish from their service browse. `dumpsys wifi` confirms `Multicast Locks held:` is empty even though the Java-side lock object reports `isHeld=true`.

`AndroidMulticastLockController` now also acquires a companion `WifiManager.WifiLock(WIFI_MODE_FULL_HIGH_PERF)` for the receiver-service lifetime. While that lock is held the device disables Wi-Fi power save, and the multicast filter Funtouch installs in power-save mode is correspondingly lifted. On stock AOSP / Pixel / Samsung the multicast lock alone already works; the Wi-Fi lock is then redundant but harmless.

⚠️ **Known limitation**: even with the WifiLock companion, **vivo Funtouch 16 / OriginOS 6 still silently refuses both locks** for non-system apps. Sustained mDNS responsiveness on vivo therefore requires either device-side configuration changes (App Freezer + battery-optimisation whitelist) or a future migration from JmDNS to Android's `NsdManager`. The vivo-specific behaviour is documented in detail in the runbook's new "Vivo / Funtouch / OriginOS specific quirks" section.

### 4. Funtouch swallows app `Log.i` — DX fix (affbcba)

CLAUDE.md already noted `OutboundConnection` uses `Log.e` to work around Funtouch's `Log.i` filter. Applied the same mitigation to the BLE advertiser, BLE scanner, and mDNS gate success-path lines (now `Log.w`). Failure-path lines were already at `W`/`E`. The events are infrequent lifecycle transitions, not log-spam, so emitting them at `W` is appropriate.

## Verification

On the Vivo X300 Ultra:

- **BLE advertise**: 167 s of clean operation, payload exactly the 24 bytes spec'd in `PROTOCOL.md`, no `DATA_TOO_LARGE`. Verified via `dumpsys bluetooth_manager`.
- **BLE scanner**: registers with `ServiceDataUuid=0000fe2c-…` and the 14-byte protocol prefix mask, scan mode flips between `BALANCED` and `LOW_LATENCY` correctly when the launcher activity goes background/foreground (Phase 5 lifecycle observer working as expected).
- **Receiver foreground service**: auto-starts from `MainActivity.onStart`, persistent notification posts, TCP listener binds.
- **mDNS gate**: override toggle propagates correctly through `MdnsVisibilityOverrideHolder` → `MdnsAdvertisementGate` → `session.publishAdvertisement()`. Initial JmDNS announce visible to a Mac on the same LAN via `dns-sd -B _FC9F5ED42C8A._tcp. local.` — service registers as `_FC9F5ED42C8A._tcp. local. <instance>` with the correct TXT and SRV records.
- **mDNS persistence on vivo**: degrades after ~30 s due to the documented Funtouch multicast-lock issue. Out of scope to fully resolve in this PR; tracked as a follow-up requiring NSD migration.

## Test plan

- [x] `./gradlew :discovery-android:testDebugUnitTest` (existing tests still green)
- [x] `./gradlew :service-android:testDebugUnitTest` (existing tests still green)
- [x] `./gradlew staticAnalysis` (clean across all five modules)
- [x] `./gradlew :app:assembleDebug` (debug APK builds)
- [x] On-device install + verify advertise broadcasts (Vivo X300 Ultra)
- [x] On-device verify scanner registration with correct ScanFilter
- [x] On-device verify receiver service auto-starts
- [x] On-device verify initial mDNS announce visible from a peer Mac
- [ ] **Future**: re-run docs/testing/interop-ble-trigger.md A1/A2 cells once a Pixel or Samsung peer is available — currently blocked because the test phone (vivo) does not ship with stock Quick Share.

### DIFF
--- a/app/src/main/kotlin/io/github/kyujincho/wvmg/MainActivity.kt
+++ b/app/src/main/kotlin/io/github/kyujincho/wvmg/MainActivity.kt
@@ -12,6 +12,7 @@ import androidx.appcompat.widget.SwitchCompat
 import io.github.kyujincho.wvmg.onboarding.PermissionRequirements
 import io.github.kyujincho.wvmg.onboarding.PermissionsOnboardingActivity
 import io.github.kyujincho.wvmg.service.receiver.MdnsVisibilityOverrideHolder
+import io.github.kyujincho.wvmg.service.receiver.ReceiverForegroundService
 
 /**
  * Empty launcher activity for the WhenVivoMeetsGoogle app.
@@ -86,13 +87,31 @@ class MainActivity : AppCompatActivity() {
         // #21 lands, the foreground service will re-check at start time
         // and surface a dismissible error instead of relaunching
         // onboarding from here.
-        if (onboardingLaunched) return
-        if (!PermissionRequirements.allGranted(this) &&
+        if (!onboardingLaunched &&
+            !PermissionRequirements.allGranted(this) &&
             !PermissionRequirements.onlyOptionalMissing(this)
         ) {
             onboardingLaunched = true
             startActivity(Intent(this, PermissionsOnboardingActivity::class.java))
+            return
         }
+
+        // Bring up the receiver foreground service so the BLE pulse
+        // scanner (#33), mDNS-publish gate (#34), and TCP listener are
+        // running while the user has the launcher visible. Without this
+        // call the production code had no other entry point that ever
+        // started the service, so the receiver pipeline was effectively
+        // dead-code on real devices — found while running the BLE-trigger
+        // interop runbook against a Vivo X300 Ultra. The service handles
+        // repeated startService calls idempotently.
+        //
+        // We only reach this point when the mandatory permissions are
+        // granted (the early-return above bounces the user to onboarding
+        // first). If the user denied an optional permission like
+        // POST_NOTIFICATIONS, BleQuickShareScanner.start() and the JmDNS
+        // publish path each re-check their own permissions internally
+        // and gracefully no-op rather than crash.
+        ReceiverForegroundService.start(this)
     }
 
     private companion object {
--- a/discovery-android/src/main/kotlin/io/github/kyujincho/wvmg/discovery/MulticastLockHolder.kt
+++ b/discovery-android/src/main/kotlin/io/github/kyujincho/wvmg/discovery/MulticastLockHolder.kt
@@ -111,7 +111,7 @@ internal class MulticastLockHolder internal constructor(
         ): MulticastLockGate {
             val wifi =
                 context.applicationContext.getSystemService(Context.WIFI_SERVICE) as WifiManager
-            val lock =
+            val multicastLock =
                 wifi.createMulticastLock(tag).apply {
                     // `setReferenceCounted(false)` makes a single release() call always
                     // fully release the lock, regardless of how many times the system
@@ -119,7 +119,26 @@ internal class MulticastLockHolder internal constructor(
                     // gives unambiguous acquire/release semantics.
                     setReferenceCounted(false)
                 }
-            return SystemMulticastLockGate(lock)
+            // Companion `WIFI_MODE_FULL_HIGH_PERF` Wi-Fi lock — the same
+            // workaround `AndroidMulticastLockController` uses on the
+            // receiver side. vivo / Funtouch / OriginOS silently refuses
+            // multicast-lock acquires for non-system apps; while the
+            // HIGH_PERF Wi-Fi lock is held, the device disables Wi-Fi
+            // power save and the multicast filter Funtouch installs in
+            // power save is correspondingly lifted. On stock AOSP this
+            // is redundant but harmless. Without this, the sender-side
+            // mDNS browse on vivo silently sees zero peers — found
+            // while running docs/testing/interop-ble-trigger.md against
+            // a Vivo X300 Ultra peering with a Galaxy S24 Ultra.
+            val wifiLock =
+                wifi
+                    .createWifiLock(
+                        WifiManager.WIFI_MODE_FULL_HIGH_PERF,
+                        "$tag-highperf",
+                    ).apply {
+                        setReferenceCounted(false)
+                    }
+            return SystemMulticastLockGate(multicastLock, wifiLock)
         }
     }
 }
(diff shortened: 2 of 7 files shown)


kyujin-cho:
## Update from extended on-device testing (Vivo X300 Ultra + Galaxy S24 Ultra)

Two more commits added to this branch (`02b9420`, `60bbeec`).

### What works after fixes 1–4 (`607c41e` + `81863dd` + `e57b0ae` + `affbcba`)

Verified on real hardware:

- **BLE advertise**: Vivo (sender) emits the 24-byte service-data pulse with no `DATA_TOO_LARGE`. Confirmed via Vivo's `dumpsys bluetooth_manager` AND via the Samsung receiver's `nearby_sharing` BLE scan filter (`Results=19` for `0xfe2c FC 12 8E`).
- **Samsung-side detection**: Samsung's NearbySharing internal logs `EndpointDiscovered(endpointId=…)` within **~2 s** of our advertise start — well under the 5 s acceptance criterion in `interop-ble-trigger.md`. PR #89's send-side ConnectionRequest fixes are observable downstream of this point.
- **Receiver auto-start** on launcher open, **lifecycle-driven scan-mode flip**, **mDNS gate publish on override toggle**: all observed working.

### Newly identified — sender-side mDNS browse blocked on vivo

After the BLE half lands cleanly, the next step in the protocol is for the WVMG sender's mDNS browse to discover the receiver's `_FC9F5ED42C8A._tcp.local.` advertisement, resolve to its IP+port, open a TCP connection, and send `ConnectionRequestFrame`. On the Vivo X300 Ultra (Funtouch 16 / OriginOS 6) this step **fails silently**: the Vivo's WVMG peer picker shows zero peers even when:

- the Samsung's Quick Share is publishing on the LAN (`dns-sd -B _FC9F5ED42C8A._tcp. local.` from a third-party Mac sees `Add 14 _FC9F5ED42C8A._tcp.local. <inst>` instantly),
- the Vivo's WVMG is on the same Wi-Fi SSID (`/24` ping reachable in both directions),
- the Vivo's WVMG has `WifiManager.MulticastLock` AND `WifiManager.WifiLock(WIFI_MODE_FULL_HIGH_PERF)` ostensibly held (Java-side `isHeld` returns `true`),
- and the Vivo's app has a UDP listener on port 5353 (`/proc/net/udp6 awk '$8 == 10333 && $4 == "07"'` shows the binding).

The smoking gun is in `dumpsys wifi`:

```
Multicast Locks held:           ← (empty)
Active lock owners: {}          ← our HIGH_PERF lock not honoured
…
Wlan Wake Reasons: … rxMulticast 114 … ipv4RxMulticast 0 …
```

vivo's Wi-Fi radio layer is **silently dropping inbound IPv4 multicast** for non-system apps despite the lock APIs reporting success — same root cause as the publish-side issue, just affecting browse too.

The proper fix is a discovery-layer migration from JmDNS to Android's built-in `NsdManager`, which uses the system mDNS responder process (which has the multicast filter exemption baked in). That is a meaningful architectural change and is out of scope for this PR. Documented as a known device limitation in `docs/testing/interop-ble-trigger.md`.

### Practical matrix status

| Cell | Status with this PR |
|------|---------------------|
| A1 (Pixel receiver) | **Blocked** with vivo as sender — TCP step never lands. Re-test once non-vivo sender available. |
| A2 (Samsung receiver) | **BLE half passes** (`EndpointDiscovered ≤ 2 s`); **mDNS half blocked** by vivo radio filter. |
| A3 (mobile-data-only) | Same as A1/A2 — blocked behind the browse step. |
| A4 (screen-off) | Same. BLE pulse delivery itself unaffected — Samsung wakes from screen-off as designed. |
| A5 (power) | Skipped — needs Pixel 7. |

### Recommended next steps

1. Land the four fixes in this PR — they all stand on their own merit (`DATA_TOO_LARGE` is universal; auto-start affects every device; the WifiLock companions are the right thing on stock AOSP / Pixel / Samsung; the Log.w upgrade is a Funtouch DX win).
2. Open a follow-up issue: **migrate `:discovery-android` from JmDNS to `android.net.nsd.NsdManager`**. This is the single change that would make the matrix runnable end-to-end on vivo.
3. Re-run the matrix from a Pixel sender (when available) to confirm A1 passes; the receiver-side fixes are already validated by Samsung's `EndpointDiscovered` log line during this session.

Issue #100, opened 2026-04-28
test(discovery): add Robolectric integration coverage for AndroidNsdBrowser / AndroidNsdRegistrar


kyujin-cho:
## Background

PR #99 (issue #98) migrated the mDNS layer from JmDNS to NsdManager.
The JVM unit tests cover the timeout / queue primitives and the
attribute byte-pass-through, but none of them exercise an actual
`NsdManager` call path — the AGP unit-test stub jar throws
`ClassFormatError` when a test allocates `NsdServiceInfo`, which forces
every test that needs a real `NsdServiceInfo` to use a seam lambda.

This was noted in the finalization pass of #99 as a deliberate
out-of-scope item: adding Robolectric (or instrumented `androidTest`)
coverage requires its own setup investment.

## What this issue covers

1. Add Robolectric to `:discovery-android` and configure a shadow for
   `NsdManager` so that `AndroidNsdBrowser.runResolve` can be exercised
   end-to-end (including the single-flight queue draining correctly).
2. Cover the reflective `setBinaryAttribute` path on a pre-API-33
   Robolectric runtime — verify the `InvocationTargetException`
   catch branch introduced in #99 is reachable.
3. Cover `AndroidNsdRegistrar.register` happy-path and registration-failed
   callbacks against a fake `NsdManager` shadow.

## Acceptance criteria

- [ ] Robolectric tests run as part of `:discovery-android:testDebugUnitTest` (no separate instrumented runner needed).
- [ ] `AndroidNsdBrowser` single-flight queue and timeout tested with a
  `NsdManager` shadow that can simulate a hanging or failing resolve.
- [ ] `AndroidNsdRegistrar` happy-path (`onServiceRegistered`) and
  failure path (`onRegistrationFailed`) covered.
- [ ] CI (`.github/workflows/ci.yml`) passes with the new tests included.

## Relates to

- PR #99 (NsdManager migration)
- Issue #98

kyujin-cho:
## Partial address via PR #102

PR #102 ([test(discovery): JVM regression guard for f1bdcb5 pre-resolve detection](https://github.com/kyujin-cho/WhenVivoMeetsGoogle/pull/102)) has been finalised and is ready to merge. Here is a summary of what it delivers and what remains deferred.

### What PR #102 delivers

[...]

PR #102, opened 2026-04-28
test(discovery): JVM regression guard for f1bdcb5 pre-resolve detection (#100)

kyujin-cho:
Partially addresses #100. Depends on PR #99.

## Summary

Adds the smallest-scope regression guard for the f1bdcb5 fix from PR #99 — the `onServiceFound` pre-resolve detection that prevents the WVMG sender's peer picker from staying empty on every modern Android device. Lifts the post-resolve fast-path predicate into a JVM-friendly companion helper `AndroidNsdBrowser.shouldSkipResolve(port: Int)` so a unit test can pin the contract directly, without requiring an `NsdServiceInfo` allocation (which trips `ClassFormatError` on the AGP unit-test stub jar).

Three new tests in `AndroidNsdBrowserTest`:

- `shouldSkipResolve returns true for a positive port` — pins true for the ports actually observed during on-device verification (32861 on Vivo, 53601 on Samsung) plus the port-range edges (1, 65535).
- `shouldSkipResolve returns false for port 0` — preserves the legacy `resolveService` round-trip on older Android (pre-MdnsDiscoveryManager pipeline always delivered port=0).
- `shouldSkipResolve returns false for negative port` — defensive for malformed records.

If a future change flips the predicate (e.g. `> 0` → `>= 0`, removes the shortcut entirely, or special-cases ports), one of these tests fails immediately.

Also keeps the test seam `AndroidNsdBrowser.resolveTimeoutMillis` (constructor parameter, default `RESOLVE_TIMEOUT_MILLIS`) added during this work — useful for any future integration test that wants to shorten the wall-clock wait without virtual time, without changing production semantics.

## Why this isn't the full Robolectric coverage #100 asked for

Issue #100's original goal was Robolectric integration coverage of `AndroidNsdBrowser` and `AndroidNsdRegistrar` so that the full `onServiceFound → resolve queue → resolveService → Resolved` flow could be exercised end-to-end on the JVM-side test runtime. Investigation surfaced a non-trivial interaction between Robolectric, the AGP unit-test stub jar, and Kotlin's lambda-compilation strategy:

1. The AGP unit-test stub jar's `NsdServiceInfo` lacks Code attributes for non-abstract methods, so the JVM throws `ClassFormatError` on any access path that loads the class.
2. Kotlin compiles captured-variable lambdas (the dominant pattern in the test file) with synthetic methods on the **outer class** whose method descriptors reference the lambda's platform-typed parameters (e.g. `NsdManager.DiscoveryListener`). Verified by `javap -v` on the compiled test class — its constant pool contains `android/net/nsd/NsdManager` directly.
3. JUnit Platform discovery (both Jupiter and Vintage engines) calls `Class.getDeclaredMethods` during the test scan. The JVM eagerly resolves the method descriptors of the outer test class, which loads `NsdServiceInfo` from the AGP stub jar, which throws `ClassFormatError` — **before** Robolectric's sandbox classloader has a chance to install its own `NsdServiceInfo` shadow.

Workarounds exist but all carry their own scope:

- Switch JUnit Platform's default `LauncherDiscoveryListener` from `abortOnFailure` to `logging` — fixes Jupiter's discovery but Vintage still aborts on the same `ClassFormatError`.
- Restructure the test class to use only reflective `NsdServiceInfo` construction (`Class.forName`-based) so its constant pool is clean of platform types — possible but ugly, makes tests hard to read.
- Enable `-Xlambdas=indy` module-wide so Kotlin emits lambdas via `invokedynamic` instead of synthetic outer-class methods — fixes the constant-pool issue but changes the lambda compilation strategy across the entire `:discovery-android` module, which has subtle effects on production code.
- Switch to Robolectric's experimental JUnit 5 extension — drops Vintage but Jupiter still trips the same `ClassFormatError` during discovery.
- Split Robolectric tests into a separate Gradle test task with a fully-isolated classpath that omits the AGP stub jar — biggest harness change but the most robust long-term path.

Each option exceeds the appropriate complexity budget for a follow-up coverage PR. **Issue #100 stays open** with this PR as a partial address — the predicate-level regression guard delivers most of the practical value (the f1bdcb5 contract is now testable on the JVM, future regressions fail fast in CI) without committing to one of the harness rewrites. Detailed notes for whoever picks #100 up next are in the commit message body.

## Branch / merge order

Branched from `feature/issue-98-nsd-migration` (PR #99), targets the same branch for review parity. Suggested merge order: **#97 → #99 → #100** (this PR). When #99 squashes into `main`, this PR's predicate extraction merges cleanly on top.

## Verification

- `./gradlew :discovery-android:testDebugUnitTest staticAnalysis :app:assembleDebug` — green.
- The three new tests run inside the existing `AndroidNsdBrowserTest` Jupiter suite. No new dependencies, no new gradle tasks, no Robolectric harness.

### DIFF
--- a/discovery-android/src/main/kotlin/io/github/kyujincho/wvmg/discovery/AndroidNsdBrowser.kt
+++ b/discovery-android/src/main/kotlin/io/github/kyujincho/wvmg/discovery/AndroidNsdBrowser.kt
@@ -62,6 +62,13 @@ internal class AndroidNsdBrowser(
             @Suppress("DEPRECATION")
             nsdManager.resolveService(info, listener)
         },
+    /**
+     * How long to wait for a `resolveService` callback before emitting
+     * an [NsdBrowserEvent.Error] and unblocking the queue. Defaults to
+     * [RESOLVE_TIMEOUT_MILLIS]. Overridable by tests to shorten the
+     * real-wall-clock wait in integration tests without virtual time.
+     */
+    internal val resolveTimeoutMillis: Long = RESOLVE_TIMEOUT_MILLIS,
 ) : NsdBrowser {
     override fun discover(serviceType: String): Flow<NsdBrowserEvent> =
         callbackFlow {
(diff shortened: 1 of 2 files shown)


kyujin-cho:
Auto-closed when base branch `feature/issue-98-nsd-migration` was deleted as part of merging #99. Reopened as **#103** with the same single commit rebased onto `main`. Full review history (clean pr-reviewer / pr-security-checker / pr-finalizer cycles, no findings) carried over.

PR #108, opened 2026-04-28
fix(protocol): Galaxy Quick Share file transfer — FILE two-frame quirk + safe-disconnect handshake

kyujin-cho:
## Summary

Fixes "couldn't receive file" on the Galaxy when sending from this app over Wi-Fi LAN. Verified on sender vivo PD2547 → receiver Samsung Galaxy S26 Ultra (One UI 8.0.5 / Android 16). Pre-fix the receiver UI showed `couldn't receive file. <sender> has been notified and can send again` with `RECEIVE_PAYLOAD_FAILED [NULL_MESSAGE]` at NearbyMediums and no file on disk, even though UKEY2 + ConnectionResponse + the PKE/PKR/Introduction Sharing-frame negotiation all completed cleanly.

Three independent receiver-side requirements were not being met:

- **FILE payload two-frame shape.** Stock Quick Share on Samsung One UI 7+ rejects FILE chunks that fuse the body with the `LAST_CHUNK` flag. `PayloadTransferEncoder.encodeFilePayload` now matches the BYTES two-frame shape: data chunks all carry `flags = 0`, followed by a dedicated empty-body terminator with `flags = LAST_CHUNK` at `offset = total_size`. NearDrop's macOS receiver tolerates the fused shape, which masked the bug in earlier interop testing.
- **`FileMetadata.id` + `IntroductionFrame.use_case`.** Stock senders set `FileMetadata.id` (attachment uuid; reusing `payload_id` satisfies "unique across all attachments") and `IntroductionFrame.use_case = NEARBY_SHARE`. Without `id`, Samsung's sharing layer cannot reconcile the inbound FILE payload's `payload_id` to the announced metadata and silently discards the attachment.
- **Safe-disconnect handshake.** PR #106 added `safe_to_disconnect_version = 1` to our `ConnectionResponseFrame` to clear an earlier silent-FIN; our terminal `DisconnectionFrame` was the empty default though. When both peers have advertised v1, Samsung's NC enforces the matching teardown handshake on every disconnect; an abrupt FIN before its read pipeline drains makes it mark every in-flight payload as failed. Every outbound `DisconnectionFrame` now carries `request_safe_to_disconnect = true`, and the orchestrator drains the wire for up to 1500 ms after the dispatch loop returns, waiting for the peer's `ack_safe_to_disconnect = true` (or peer FIN) before closing. Verified on-device: peer acknowledges within ~120 ms.

Plus diagnostic instrumentation that should have been there from day one:

- `OutboundConnectionDriver` now logs every dispatch event, every FSM effect, the `streamFilesAndComplete` lifecycle (start, per-file boundaries, first-chunk-written, loop-end, Disconnection), and the terminal-result resolution. Same `Log.e` + on-device file-fallback path the existing step 1–8 logging uses (vivo Funtouch swallows `Log.i`). Without these, the dispatch loop's silence past `step 8: sending initial PKE frame` made the bug invisible.

Cleanup:

- `SendActivity` had pre-existing `staticAnalysis` violations from #106 (its CI never validated due to the billing issue) — extracted `upsertResolvedPeer` to flatten `onDiscoveryEvent`'s nesting and added a kdoc-rationaled `@Suppress("TooManyFunctions")` on the class.

## Test plan

- [x] `./gradlew :core-protocol:test` — full JVM suite green, including `PayloadTransferEncoderTest` regression for the single-chunk small-file two-frame shape.
- [x] `./gradlew staticAnalysis` — ktlint + detekt clean across all modules.
- [x] `./gradlew :app:assembleDebug` — debug APK builds.
- [x] On-device interop: vivo PD2547 → Galaxy S26 Ultra (One UI 8.0.5), 89 KiB JPEG and 226 KiB PNG. Pre-fix: receiver UI shows "couldn't receive file" 100% of the time. Post-fix: file lands on disk; receiver shows the standard "received" path; safe-disconnect ack handshake completes (`fsm: safe-disconnect peer Disconnection ack=true` in on-device log).


kyujin-cho:
## PR Finalization Complete

### Summary

- **Tests**: Already complete (IntroductionFramesTest + PayloadTransferEncoderTest regression guard both green). No new tests added — existing coverage is thorough.
- **Lint/Format**: staticAnalysis clean (ktlint + detekt). No fixes needed.
- **CI**: `.github/workflows/ci.yml` unchanged — no updates required.
- **Documentation**: Updated in commit `5010530` on `bugfix/galaxy-file-transfer-quirks`.

### Documentation changes (`docs:` commit)

**`CLAUDE.md` — three new "Quick Share interop quirks worth knowing" entries:**
- FILE payloads require a dedicated empty `LAST_CHUNK` terminator (Samsung One UI 7+)
- Outbound `DisconnectionFrame` must set `request_safe_to_disconnect=true` with 1500 ms wire drain, gated on the terminal type
- `IntroductionFrame.use_case=NEARBY_SHARE` and non-zero `FileMetadata.id` required by One UI 8.0.5

**`README.md` — payload layer description clarified:** the two-frame terminator shape now explicitly covers FILE payloads in addition to BYTES, with a one-sentence note about the Samsung One UI 7+ requirement.

**`docs/testing/interop-stock-quick-share-android.md`:**
- Test matrix section updated to explain why a small-file (single-chunk) run is mandatory alongside large-file: it's the regression guard for the fused-terminator failure mode
- Three new Samsung / One UI quirk entries for the PR #108 fixes
- New "Samsung-specific checks" subsection under Test 2 procedure with logcat confirmation steps (`fsm: safe-disconnect peer Disconnection ack=true`) and the single-chunk file integrity check

All checks passing. Ready for merge.

PR #134, opened 2026-04-29
fix: sequence outbound discovery after receiver unpublish (#107)

### LINKED ISSUE #107: discovery(funtouch): browse listener stops returning peers once we register a service of the same type
## Summary

On a Vivo X300 Ultra running Funtouch (Android 16), once `Discovery.advertise` registers a service of type `_FC9F5ED42C8A._tcp` via `NsdManager.registerService`, a concurrently-running `NsdManager.discoverServices` listener for the *same* service type stops surfacing new peers — and does not recover even after the advertise is unregistered.

The user-visible symptom: with the "Always visible" toggle off, the picker correctly shows the Galaxy S24 Ultra peer; the moment the toggle is flipped on (which starts the receiver-side mDNS publish), the picker goes empty, and toggling back off does not restore visibility for the rest of the app session.

## Reproduce

1. WVMG built from `bugfix/galaxy-handshake-and-picker` (PR #106) installed on a Vivo X300 Ultra on Funtouch / Android 16. Galaxy S24 Ultra peer running stock Quick Share with visibility = Everyone, both phones on the same Wi-Fi.
2. Force-stop the WVMG app.
3. Trigger a fresh share intent on the Vivo (Files / Photos → Share → WhenVivoMeetsGoogle). `SendActivity` opens. **Galaxy appears in the picker** (verified — `browse: serviceResolved name=I0NXU0L8n14AAA addrs=172.17.142.189 ...` and `browse: serviceResolved name=I1FDU0b8n14AAA ... n="Iri1...Kyujin's S26 Ultra..."`).
4. Switch to MainActivity → toggle **Always visible ON**. Receiver service publishes its mDNS record (`advertise: registered instance=I2lzSmH8n14AAA port=38127`).
5. Switch back to the share intent / SendActivity. **Picker is empty.** Discovery listener has stopped surfacing new `serviceFound` events for any peer including Galaxy, even though the system NSD's internal cache still contains Galaxy (`dumpsys servicediscovery` reports `foundServices=2 lostServices=0 fromCache=true` against our DiscoveryListener).
6. Toggle Always Visible **off**, wait through the gate's 30 s debounce until `advertise: closing handle` fires. **Galaxy still does not reappear** in the picker — the listener never re-emits Galaxy's record. A fresh process restart is required to recover.

## Logs (from PR #106's verbose discovery instrumentation)

Order of events from the failing run:

```
browse: starting collector
browse: serviceFound name=I0NXU0L8n14AAA           ← Galaxy anonymous, found
browse: serviceResolved name=I0NXU0L8n14AAA addrs=172.17.142.189 ...
browse: serviceFound name=I1FDU0b8n14AAA           ← Galaxy named, found
browse: serviceResolved ... n="Iri1...Kyujin's S26 Ultra..."
browse: collector closing                          ← user navigates to MainActivity

advertise: starting publish instance=I2lzSmH8n14AAA port=38127   ← Always Visible toggled ON
advertise: registered ...
WvmgMdnsGate: publish: published mDNS

browse: starting collector                         ← user returns to SendActivity
… no further serviceFound / serviceResolved events for any peer …
discovery snapshot: advertising=true browsing=false
```

The `Discovery/DiscoveryListener` row in `dumpsys servicediscovery` confirms the listener is registered and the system NSD's cache holds 2 services (Galaxy's anonymous + Galaxy's named). Our app's listener-callback chain is silently no longer being driven.

## Hypothesis

Vivo Funtouch's `NsdManager` implementation has a same-process / same-service-type interaction bug: once a `registerService` for a given service type lands in the same client UID that has an active `discoverServices` listener for that same service type, the listener loses its connection to the system mDNS event stream. The platform never re-binds it, even when the registered service is later torn down via `unregisterService`. This is in the same family of vendor-NSD quirks that drove the JmDNS → NsdManager migration (#98) but in the other direction.

## Acceptance criterion

Toggling "Always visible" on while WVMG is foregrounded must not permanently break peer discovery. After Always Visible is toggled off (and the gate's debounce expires), the picker must re-acquire the Galaxy peer within ~5 s.

## Candidate fixes to evaluate

1. **Share a single `Discovery` instance** across the receiver service and `SendActivity` so both register and discover happen through one NSD client object. May or may not change the underlying platform behaviour, but worth measuring.
2. **Hard-recreate `NsdManager` on register-discover conflict**: when SendActivity opens and the receiver is currently advertising, fully unregister, then `Context.getSystemService(NSD_SERVICE)` to get a fresh `NsdManager` reference (or restart the receiver service), then start the browse. Some Android quirks heal across NsdManager re-references.
3. **Mutual exclusion**: don't allow the receiver-side mDNS publish to be active while `SendActivity` is foreground at all (today the gate veto already does this immediately, but the ordering between the synchronous unregister and the discover restart is racy — measure whether sequencing them deterministically inside SendActivity.onCreate before launching the browse Flow recovers the listener).

## Scope notes

- Failing platform: Vivo X300 Ultra (V2547A), Funtouch / OriginOS / Android 16.
- Working platforms: this issue does **not** reproduce on Pixel devices in our prior interop runs (#30 / #36 testing).
- This is **not** a regression introduced by PR #106 — the underlying NSD quirk pre-dates the gate-veto code. The issue surfaced because PR #106's verbose logging makes it visible, and the gate-veto was specifically meant to mitigate it but isn't enough on its own.

## Linked context

- PR #106 (`bugfix/galaxy-handshake-and-picker`) — references this issue in the "Known follow-up" section of its body.
- #98 (NsdManager migration) — same family of vendor-NSD quirks.

kyujin-cho:
## Summary
- gate SendActivity discovery startup on receiver mDNS unpublish so vivo/Funtouch does not browse while the same process still advertises the Quick Share service type
- mirror receiver advertisement state through a process-wide holder and wait for Android NSD unregister callbacks when closing registrations
- document the #107 vivo verification path and add JVM coverage for the new holder/gate behavior

## Verification
- ./gradlew :discovery-android:testDebugUnitTest :service-android:testDebugUnitTest :app:testDebugUnitTest
- ./gradlew staticAnalysis

Closes #107

### DIFF
--- a/app/src/main/kotlin/io/github/kyujincho/wvmg/send/SendActivity.kt
+++ b/app/src/main/kotlin/io/github/kyujincho/wvmg/send/SendActivity.kt
@@ -32,6 +32,7 @@ import io.github.kyujincho.wvmg.protocol.connection.OutboundConnectionState
 import io.github.kyujincho.wvmg.protocol.endpoint.DeviceType
 import io.github.kyujincho.wvmg.protocol.endpoint.EndpointInfo
 import io.github.kyujincho.wvmg.service.receiver.OutboundSessionActiveHolder
+import io.github.kyujincho.wvmg.service.receiver.ReceiverAdvertisementStateHolder
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
@@ -80,6 +81,7 @@ public class SendActivity : AppCompatActivity() {
     private var files: List<FileSource> = emptyList()
     private val peers: MutableList<DiscoveredService> = mutableListOf()
     private val pendingPeerRemovalJobs: MutableMap<String, Job> = mutableMapOf()
+    private var outboundPresenceJob: Job? = null
     private var discoveryJob: Job? = null
     private var connectionJob: Job? = null
     private var activeConnection: OutboundConnection? = null
@@ -145,9 +147,7 @@ public class SendActivity : AppCompatActivity() {
 
         binding.sendPayloadSummary.text = PayloadSummary.forFiles(this, files)
         binding.sendSubtitle.setText(R.string.send_subtitle_discovering)
-        startDiscovery()
-        startEmptyPeerHintTimer()
-        startBleAdvertise()
+        startOutboundPresence()
     }
 
     /**
@@ -207,6 +207,7 @@ public class SendActivity : AppCompatActivity() {
         // OutboundConnection so a CancelFrame is sent on the wire when
         // possible (best-effort — already-terminal states are no-ops).
         activeConnection?.cancel()
+        outboundPresenceJob?.cancel()
         discoveryJob?.cancel()
         connectionJob?.cancel()
         emptyPeerHintJob?.cancel()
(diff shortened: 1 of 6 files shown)


kyujin-cho:
Local verification completed because GitHub Actions could not start the runner job due account billing/spending-limit configuration, before checkout or Gradle execution. Commands run locally:

- ./gradlew :discovery-android:testDebugUnitTest :service-android:testDebugUnitTest :app:testDebugUnitTest
- ./gradlew staticAnalysis
- ./gradlew :core-protocol:test :app:assembleDebug

PR #135, opened 2026-04-29
feat: reorder dual-channel upgrade frames (#133)

### LINKED ISSUE #133: feat: Implement dual-channel bandwidth upgrade receive/reorder path
## Problem / Background
Real-device testing from Galaxy S26 Ultra stock Quick Share to vivo X300 Ultra WVMG exposed a gap in WVMG's Phase 4 Wi-Fi Direct receive path. When WVMG attempted a server-side Wi-Fi Direct bandwidth upgrade, the Galaxy UI stalled at Connecting.

Logs showed Samsung sending Nearby Share application payload frames on the original encrypted Wi-Fi LAN channel while it was also processing WVMG's `UPGRADE_PATH_AVAILABLE` and switching toward `ENCRYPTED_WIFI_DIRECT`. WVMG currently assumes a single active `SecureChannel` receive path with strict global `SecureMessage` receive sequence numbers. When application frames continued arriving on the old channel while WVMG switched its active reader to the new channel, the session failed with `SequenceNumberMismatchException` (for example: expected 8, got 7).

A conservative mitigation was verified: if inbound sees an initial application frame already buffered on Wi-Fi LAN, it skips the upgrade and stays on Wi-Fi LAN. That restored stable Galaxy to vivo transfers, including a 215,942,159-byte APK transfer completed over Wi-Fi LAN. Basic functionality is therefore stable, but it does not validate or complete true Wi-Fi Direct bandwidth-upgrade behavior for stock Android and Samsung peers.

## Proposed Solution
Implement the full dual-channel bandwidth-upgrade handoff for inbound receive paths so WVMG can keep both the prior encrypted Wi-Fi LAN channel and the upgraded direct-medium channel alive during the upgrade boundary.

The implementation should:
- Treat `SecureMessage` receive sequence numbers as global across concurrently live transports during handoff, rather than binding the active receive cursor to exactly one socket at a time.
- Drain the prior and upgraded transports concurrently, reorder decrypted `OfflineFrame`s by `SecureMessage` receive sequence number, and only deliver frames to the sharing FSM when the next expected sequence is available.
- Continue handling application frames, keep-alives, cancellation and disconnection frames, and bandwidth-upgrade control frames correctly while both channels are live.
- Honor `LAST_WRITE_TO_PRIOR_CHANNEL` and `SAFE_TO_CLOSE_PRIOR_CHANNEL` semantics before tearing down the old channel.
- Preserve the conservative Wi-Fi LAN fallback path as a guardrail until the true dual-channel upgrade path is proven green in tests and on devices.
- Add JVM coverage for interleaved delivery around the upgrade boundary and update manual-device checklists to distinguish fallback success from a real upgraded-medium success case.

## Acceptance Criteria
- [ ] WVMG can accept stock Samsung/Android bandwidth upgrade to `WIFI_DIRECT` without stalling or falling back when the peer starts interleaving frames.
- [ ] The receive pipeline can drain old and upgraded transports concurrently during the bandwidth-upgrade handoff and deliver `OfflineFrame`s to the sharing FSM in `SecureMessage` sequence order.
- [ ] App frames, keep-alives, cancel/disconnection frames, and bandwidth-upgrade control frames are handled correctly while both channels are live.
- [ ] `LAST_WRITE_TO_PRIOR_CHANNEL` and `SAFE_TO_CLOSE_PRIOR_CHANNEL` semantics are honored without closing the old channel before all lower-sequence frames are consumed.
- [ ] No `SequenceNumberMismatchException` occurs when Samsung sends app frames on the old channel while the upgraded channel is established.
- [ ] Add JVM tests that simulate interleaved old/new channel delivery around the upgrade boundary.
- [ ] Update the real-device testing checklist to distinguish Wi-Fi LAN fallback success from true `WIFI_DIRECT` upgraded-medium success.

## Technical Considerations
- Keep protocol sequencing and reorder logic in JVM-testable layers (`:core-protocol`) and keep Android medium adapters in Android-only modules.
- Review whether the current `SecureChannel` / connection-driver assumptions need per-transport receive pumps plus a higher-level merge/reorder stage, instead of a single active blocking read loop.
- Preserve existing safe-disconnect and teardown invariants so the old transport is not closed before its lower-sequence frames are drained.
- Treat this as follow-up tracking for the Phase 4 medium work in #4 and the completed Wi-Fi Direct groundwork in #49; update #54's manual/integration coverage once true upgraded-medium success is testable.

kyujin-cho:
## Summary
- add a dual-channel upgrade reader that drains prior and upgraded SecureMessage transports and delivers frames in global sequence order
- let inbound upgrade proceed even when a first application frame was already buffered on Wi-Fi LAN
- add inbound diagnostic logging and update medium integration docs to distinguish true upgraded-medium success from fallback

## Verification
- ./gradlew :core-protocol:test --tests '*DualChannelFrameReaderTest'
- ./gradlew :core-protocol:test
- ./gradlew staticAnalysis
- ./gradlew :service-android:testDebugUnitTest :app:testDebugUnitTest :app:assembleDebug
- ./gradlew check (fails in pre-existing discovery-android lint: WifiP2pConfig.Builder requires API 29 while minSdk is 24)

Closes #133

### DIFF
--- a/core-protocol/src/main/kotlin/io/github/kyujincho/wvmg/protocol/connection/BandwidthUpgradeOrchestrator.kt
+++ b/core-protocol/src/main/kotlin/io/github/kyujincho/wvmg/protocol/connection/BandwidthUpgradeOrchestrator.kt
@@ -9,6 +9,7 @@ import com.google.location.nearby.connections.proto.OfflineWireFormatsProto.Band
 import com.google.location.nearby.connections.proto.OfflineWireFormatsProto.OfflineFrame
 import com.google.location.nearby.connections.proto.OfflineWireFormatsProto.V1Frame
 import io.github.kyujincho.wvmg.protocol.crypto.securemessage.SecureChannel
+import io.github.kyujincho.wvmg.protocol.crypto.securemessage.SequencedOfflineFrame
 import io.github.kyujincho.wvmg.protocol.medium.Medium
 import io.github.kyujincho.wvmg.protocol.medium.MediumRegistry
 import io.github.kyujincho.wvmg.protocol.medium.PreparedUpgradeSelection
@@ -17,6 +18,7 @@ import io.github.kyujincho.wvmg.protocol.medium.UpgradedTransport
 import io.github.kyujincho.wvmg.protocol.medium.asFramedConnection
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.withTimeoutOrNull
+import java.util.TreeMap
 
 /**
  * Blocking bandwidth-upgrade handshake runner.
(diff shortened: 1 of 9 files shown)


kyujin-cho:
Local verification for #133:

- `./gradlew :core-protocol:test --tests '*DualChannelFrameReaderTest'` passed
- `./gradlew :core-protocol:test` passed
- `./gradlew staticAnalysis` passed
- `./gradlew :service-android:testDebugUnitTest :app:testDebugUnitTest :app:assembleDebug` passed

GitHub Actions did not start the job because the account billing/spending limit is blocking runners. I also ran `./gradlew check`; it reached lint and failed on the existing `discovery-android/src/main/kotlin/io/github/kyujincho/wvmg/discovery/medium/WifiDirectGroupController.kt` NewApi finding for `WifiP2pConfig.Builder` requiring API 29 while minSdk is 24.

Issue #145, opened 2026-05-01
fix: Investigate Galaxy to vivo duplicate FEF3 BLE bootstrap failure


kyujin-cho:
## Problem / Background

PR #144 / issue #143 added BLE GATT sender bootstrap and receiver connectible fixes, but actual-device validation on **2026-05-02** found a remaining **Galaxy S26 Ultra -> vivo X300 Ultra** receiver-side BLE GATT failure.

Test setup:
- Galaxy S26 Ultra sender using Samsung ShareLive / stock Quick Share
- vivo X300 Ultra receiver running the WVMG debug APK

Observed facts:
- The WVMG receiver is connectible and discoverable. Logs showed `mAdvertiseConnectable=true`, `BLE GATT bootstrap active service=0000fef3-0000-1000-8000-00805f9b34fb`, and Samsung ShareLive discovered the `WhenVivoMeetsGoogle` endpoint `P0ZK`.
- Samsung attempted BLE bootstrap: `requestConnection() ... endpoint P0ZK over mediums [BLE]` and `successfully created a BLE EndpointChannel to endpoint P0ZK`.
- The transfer still failed before receiver consent with `Failed to write ConnectionRequestFrame` and `BleSocketOutputStream failed to write data`.
- The vivo WVMG log only saw GATT slot reads and never saw socket CCCD or write callbacks.
- Samsung's GATT cache showed two local `0000fef3-0000-1000-8000-00805f9b34fb` services on the vivo: a platform socket-only service first, and WVMG's slot-bearing service second.

Current hypothesis: stock GMS reads WVMG's slot from the second `0xFEF3` service but opens the BLE socket against the first duplicate platform `0xFEF3` service, so the Nearby socket never reaches the WVMG receiver.

## Proposed Solution

- Reproduce the Galaxy -> vivo failure with the same actual-device loop and confirm whether duplicate local `0xFEF3` services are the decisive cause.
- Investigate a viable coexistence strategy for WVMG and the platform Quick Share stack on the same Android receiver, including whether WVMG can avoid or isolate the duplicate-service collision while still advertising a stock-compatible `0xFEF3` path.
- If Android / GMS cannot reliably disambiguate duplicate local `0xFEF3` services, document that as a firm platform limitation and define the expected product behavior or fallback.
- Add targeted diagnostics around service-instance discovery, slot reads, socket open attempts, CCCD writes, and characteristic writes so future device validation can prove which service instance was selected.

## Acceptance Criteria

[...]

PR #152, opened 2026-05-04
feat(consent): show foreground modal when LibreDrop is resumed

### LINKED ISSUE #151: feat: Show in-app consent modal instead of relying on the heads-up notification when LibreDrop is in the foreground
## Problem / Background

LibreDrop's receiver currently surfaces incoming-transfer consent purely as a high-priority `category=call` heads-up notification with `Accept` / `Reject` action buttons. This is the right surface when the app is backgrounded, but it is awkward when the user is already looking at LibreDrop's `MainActivity`:

- On some OEMs the heads-up is throttled or swallowed when the launcher is foregrounded (vivo Funtouch and Samsung One UI both have variants of this), so the user sees nothing happen even though `InboundConnectionState` is `WaitingForUserConsent`.
- Even when the heads-up does appear, it pulls focus into the notification shade for a flow the user is already staring at — strictly worse than an in-activity modal.
- During the #145 hardware loop the consent notification's `Accept` action could not be reached from the shade on the Vivo X300 Ultra without firing the broadcast manually; an in-activity prompt would have been the natural surface.

There is already a `ConsentIntents.ACTION_SHOW_CONSENT` action and a `activity_consent_trampoline.xml` layout (see the consent package: `service-android/src/main/kotlin/dev/bluehouse/libredrop/service/receiver/consent/`), which today wires the notification's body tap to a trampoline that renders the consent UI in-process. That trampoline is the natural building block for the foreground path.

## Proposed Solution

When `MainActivity` (or any other LibreDrop activity that owns the consent surface) is `RESUMED` and a new `WaitingForUserConsent` state arrives:

- Show the consent UI as an in-activity modal (a `DialogFragment` over `MainActivity`, or by routing the trampoline activity as `singleTop` with `FLAG_ACTIVITY_REORDER_TO_FRONT`, whichever fits the existing `ConsentCoordinator` shape better). The dialog must show the same fields as the notification: source device name, transfer item summary, PIN, Accept / Reject actions.
- Suppress (or downgrade to `IMPORTANCE_LOW` / `silent`) the heads-up notification for the same connection id while the foreground modal is up. The notification path still has to work as a fallback for any consent that arrives while no LibreDrop activity is in the foreground, so the suppression must be **per connection id**, not a global flag.
- If the user backgrounds LibreDrop while the modal is up (without deciding), restore the heads-up notification for that connection so they can resume from the shade. Symmetrically, if a notification has already been raised and the user then foregrounds LibreDrop, dismiss the heads-up and present the modal.

## Acceptance Criteria

- [x] Foreground path: when `MainActivity` is resumed and a new inbound transfer needs consent, the user sees an in-activity modal (no heads-up notification flash, no shade interaction needed) and tapping Accept / Reject routes through the existing `ConsentCoordinator` / `ConsentBroadcastReceiver` so the inbound state machine treats the decision identically to the notification path.
- [x] Background path: when no LibreDrop activity is resumed, the existing notification path is unchanged.
- [x] Foreground/background transition: the surface follows the user — modal disappears (and notification re-raises) when the activity is paused; notification clears (and modal opens) when the activity is resumed.
- [x] Per-connection-id scoping: a notification for connection A must not be suppressed because the user is foregrounded for an unrelated previous connection B that has already been decided.
- [x] No protocol-level changes; the receiver state machine, GATT bootstrap, and bandwidth-upgrade paths are untouched.

## Technical Considerations

- Existing consent surface lives in `service-android/src/main/kotlin/dev/bluehouse/libredrop/service/receiver/consent/` — `ConsentCoordinator`, `ConsentNotification`, `ConsentBroadcastReceiver`, `ConsentIntents`, `ConsentRegistry`. The action constants `ACTION_ACCEPT`, `ACTION_REJECT`, `ACTION_SHOW_CONSENT` and the `EXTRA_CONNECTION_ID` long are already defined.
- `app/src/main/res/layout/activity_consent_trampoline.xml` is the existing in-app consent layout that the notification's `contentIntent` opens today.
- `MainActivity.onStart` already runs a few side-effect refreshes (`refreshBatteryBanner`, `refreshLegacyWvmgBanner`); a similar `refreshPendingConsent` hook could observe the consent registry for a `WaitingForUserConsent` entry and present the modal.
- The decision broadcast that the modal sends should reuse `ConsentBroadcastReceiver` rather than a parallel path, so accepted-from-foreground vs accepted-from-notification go through identical bookkeeping.

## Out of Scope

- Lock-screen visibility of consent (a separate accessibility / privacy concern).
- Multiple-pending-consents UI (current behaviour stacks them as separate notifications and will continue to; the modal version can be one-at-a-time, queueing the others as notifications).
- QR-paired / always-trusted senders.

kyujin-cho:
## Summary

Closes #151. When a \`WaitingForUserConsent\` arrives and any LibreDrop activity is foregrounded, the receiver launches the consent trampoline as an in-app modal instead of relying on the heads-up notification. The modal follows the user across foreground/background transitions:

- **Foreground arrival** -> in-app modal, no heads-up flash.
- **Background arrival** -> existing heads-up notification path, unchanged.
- **Background while modal up** -> coordinator cancels the modal and raises the notification so the user can resume from the shade.
- **Foreground while notification up** -> coordinator cancels the heads-up and launches the modal.
- **Per-connection-id scoping** -> a decided connection never has its notification resurrected by a later foreground/background flap belonging to a different connection.

Decision routing converges on \`ConsentBroadcastReceiver\`: the trampoline's Accept/Reject buttons fire \`ACTION_ACCEPT\` / \`ACTION_REJECT\` broadcasts identical to the notification action buttons, so the inbound FSM cannot tell which surface produced the decision.

No protocol-level changes - diff is confined to \`:app\` and \`:service-android\`.

## Architecture

- \`AppForegroundState\` (new, \`service-android/.../foreground/\`) - small SAM-listener interface plus a \`ProcessLifecycleOwner\`-backed adapter and a pure-JVM \`InMemoryAppForegroundState\` for tests. Wraps the lifecycle-process API so the coordinator stays Robolectric-free.
- \`ConsentCoordinator\` - now tracks per-connection surface state (\`Notification\` / \`Modal\` / \`Decided\`). Picks the launch surface from \`AppForegroundState.isForeground\` at consent arrival; subscribes to foreground transitions and switches surfaces in lock-step. \`Decided\` is a tombstone that prevents re-raising a notification for a connection the user has already responded to.
- \`Sink\` interface gains \`launchModal\` / \`dismissModal\` alongside the existing \`postConsent\` / \`dismissConsent\`.
- \`ConsentModalRegistry\` (new) - in-memory map keyed by connection id from the live trampoline activity to a \`finish()\` callback. Lets the coordinator dismiss a modal from the service scope without needing an Activity reference.
- \`ConsentTrampolineActivity\` - on Accept/Reject now broadcasts through \`ConsentBroadcastReceiver\` instead of calling \`submitUserConsent\` directly. Registers itself with \`ConsentModalRegistry\` on create, unregisters on destroy. \`onNewIntent\` rebinds when the activity gets re-launched (singleTop) under a different connection id.
- Manifest - \`ConsentTrampolineActivity\` switched from \`singleTask\` to \`singleTop\` so coordinator-driven re-launches route through \`onNewIntent\` rather than spawning a duplicate task. \`excludeFromRecents\` retained.
- \`ReceiverForegroundService\` - lazily wires a \`ProcessLifecycleOwnerAppForegroundState\` on the main thread, threads it into the consent coordinator, and adds a \`Sink\` impl whose \`launchModal\` calls \`startActivity\` on the trampoline with \`NEW_TASK | REORDER_TO_FRONT | SINGLE_TOP\` and whose \`dismissModal\` invokes \`ConsentModalRegistry.dismiss\`.

## Validation

- [x] \`./gradlew staticAnalysis\` - clean.
- [x] \`./gradlew :app:testDebugUnitTest :service-android:testDebugUnitTest :core-protocol:test\` - all green.
- [x] \`./gradlew :app:assembleDebug\` - clean.
- [x] New unit tests:
  - \`AppForegroundStateTest\` - covers initial value, transition coalescing, listener fan-out, cancel idempotence.
  - \`ConsentCoordinatorTest\` (additions) - foreground arrival launches modal, background arrival posts notification, background-while-modal swaps to notification, foreground-while-notification swaps to modal, decided connection is not re-surfaced, terminal-after-modal dismisses the modal, repeated foreground transitions while modal up are no-ops.

## Known limitations / Follow-ups

Two accepted-MEDIUM nits from pr-reviewer, intentionally deferred:

- **M1 — surface-FSM race on rapid foreground/background flips**: if the user backgrounds and reforegrounds within a single coroutine dispatch window, \`onForegroundChanged\` fires twice before the first transition's side-effects complete. The coordinator's \`synchronized(surfaceLock)\` prevents a double-launch, but the dismissed surface may briefly linger. In practice this requires sub-millisecond UI thrashing and is invisible to the user; a follow-up can introduce a debounce if real-device testing surfaces it.
- **M2 — silent BAL fallback on hostile OEMs**: \`startActivity\` with \`NEW_TASK | REORDER_TO_FRONT\` is subject to Background Activity Launch (BAL) restrictions on AOSP 10+. The coordinator falls back gracefully to the notification path when the trampoline cannot be brought forward (the activity simply won't appear), but it does so silently — no log line or metric. OEMs with aggressive BAL policies (e.g. some Xiaomi/vivo ROM variants) may always land on the notification path even when the app appears foregrounded. A follow-up can add explicit BAL-status logging to \`ReceiverForegroundService.Sink.launchModal\`.

## Notes

- \`:service-android:robolectricDebugUnitTest\` was mentioned in the prompt but does not exist in the project (only \`:discovery-android\` has the Robolectric task wired up). All new tests stay pure-JVM.
- A pre-existing \`:discovery-android:lintDebug\` failure (\`BluetoothDevice#connectGatt\` API-level warning at \`BleGattInitialControlClient.kt:196\`) is not introduced by this change and is not part of CI (\`.github/workflows/ci.yml\` runs \`staticAnalysis\` + tests + \`:app:assembleDebug\` only).


kyujin-cho:
## Hardware regression — DO NOT MERGE

Validated this PR on the **Vivo X300 Ultra (V2547A on Funtouch 16 / OriginOS 16.0.19.10.W10)** that issue #145's hardware repro lives on. The foreground-modal path **does not render on this device**. Stock Galaxy → Vivo Quick Share fails strictly worse than the pre-#152 notification-only baseline (which was validated end-to-end during the #150 debug loop).

### Repro

1. Vivo: install this PR's `app-debug.apk`, foreground `MainActivity`, toggle Always Visible on.
2. Galaxy S26 Ultra: stock Quick Share, share text to V2547A.

### Observed flow

- BLE GATT bootstrap → Wi-Fi Direct upgrade → PairedKey/Introduction all complete cleanly. `libredrop-inbound.log` reaches `WaitingForUserConsent(metadata=...)` at `1777870576772` (only ~4s after the Galaxy tap).
- `ConsentTrampolineActivity` is launched (visible in `dumpsys activity activities` as `mLastPausedActivity`), but immediately transitions to **`t=-1 d=-1 f`** (no task, finishing).
- The libredrop task (`Task{ded7add #1279}`) keeps `sz=1` throughout — the trampoline never landed in `MainActivity`'s task; it launched into a new task that was instantly torn down. `MainActivity` remains `topResumedActivity`.
- **No heads-up notification is raised either** (no `incoming_transfer` channel record in `dumpsys notification`). Only the foreground-service notification (id=`1279411282`, channel=`libredrop.receiver.foreground`) is active.
- Galaxy times out at 60s with `Cancelled(cause=PEER)`. The transfer fails and the user has had no surface to interact with.

### What this rules out

- **Lookup-null at `bindIntent`** (line 141): `observeOneConnection` calls `registry.register(connectionId, entry)` before `raiseConsentSurface`, and both reference the same `ConsentRegistry.instance` singleton. The trampoline's `lookup` should always succeed.
- **Pure FG→BG swap from `M1`**: pr-reviewer's M1 race was classified LOW because the swap raises a notification as the modal closes. Here, the modal closes but no notification is raised — so the path that fired is `dismissConsentSurface` (which marks `Surface.Decided` and only calls `dismissModal`), not `applySurfaceSwitch(target=Notification)`.

### Likely root cause (to confirm in fix work)

`dismissConsentSurface` is called once during the `WaitingForUserConsent` window via a path I haven't traced yet — most plausibly because the launch flags `FLAG_ACTIVITY_NEW_TASK | FLAG_ACTIVITY_REORDER_TO_FRONT | FLAG_ACTIVITY_SINGLE_TOP` plus `excludeFromRecents="true"` interact with Vivo Funtouch's BAL-adjacent heuristics so that the trampoline lands in a transient new task that is immediately removed. The activity record going to `t=-1` corroborates this. Whatever the exact mechanism, the activity is gone before the user can see anything, and the surface FSM ends up in `Decided` without ever raising a fallback notification.

### Recommendation

Block merge. Pick one:
- **(a) Hard fix**: ensure the trampoline lands in `MainActivity`'s task (drop `NEW_TASK` when launching from a context that's already foregrounded; or set explicit `taskAffinity` matching MainActivity; or launch from `MainActivity` itself via a `PendingIntent` callback). Then validate on Vivo Funtouch + Pixel + Samsung One UI.
- **(b) Defensive fallback**: when `dismissConsentSurface` would mark a connection `Decided` *while it's still in `WaitingForUserConsent`* (i.e., the protocol-side state hasn't reached a terminal), re-raise the heads-up notification instead of going silent. That would have salvaged this run on Vivo even with the trampoline failing to render.
- **(c) Roll back to notification-only** (keep #150 baseline) and re-attempt #151 with the lessons learned.

I'd suggest (b) as a guardrail regardless of (a) — silent surface loss is the worst possible failure mode for a consent prompt.

### Test bench notes

Galaxy's stock Quick Share aggressively caches per-peer session state. After a cancelled send, a second `am start ... SEND` to the same Sharelive activity opens the share sheet showing the prior peer in a "Failed" state but does not re-send when the peer is tapped — `force-stop com.samsung.android.app.sharelive` is required between attempts to get a clean negative test. This is unrelated to LibreDrop but worth documenting for whoever picks up the fix.

### Unaffected by this regression

- Pre-#152 notification-only path (validated #150).
- BLE GATT bootstrap, Wi-Fi Direct upgrade, sharing FSM, payload delivery — none touched by this PR.
- The migration banner from PR #150.

kyujin-cho:
## Hardware regression — diagnosed and fixed (2 commits)

**TL;DR:** Foreground modal works on Vivo Funtouch 16. The earlier "trampoline finishes immediately" symptom was actually two separate problems: (1) test-bench state contamination from a stale prior cancellation in the same process, and (2) a real bug in the modal-Accept broadcast routing. Both unblocked by the new commits.

### Root cause

`ConsentTrampolineActivity.submit` dispatched Accept / Reject through:

```kotlin
sendBroadcast(Intent(action).apply {
    setPackage(packageName)
    setClass(this, ConsentBroadcastReceiver::class.java)   // <-- the bug
    putExtra(EXTRA_CONNECTION_ID, connectionId)
})
```

`ConsentBroadcastReceiver` is registered **dynamically** by `ReceiverForegroundService.registerConsentReceiver()` with `RECEIVER_NOT_EXPORTED` — there is no manifest entry. `setClass(...)` binds the broadcast to a manifest-registered component name, and the BroadcastQueue silently drops it when no manifest entry resolves.

Pre-merge tests for #150 only reached this path via `am broadcast -a ... -p ...` (action + package only, no `setClass`), which the dynamic receiver does match. The trampoline's own `submit` path was never end-to-end tested before, so the bug shipped.

### Diagnostic trace that nailed it

`7ee0816` adds a `ConsentDiagnostic` appender that writes to the existing `libredrop-inbound.log` file (Funtouch silently drops `Log.i` / `Log.w` from non-vendor-blessed apps, so the on-disk fallback is the reliable channel). With the trace in place, the failing run on the unfixed build looked like:

```text
1777871997759 consent: raise id=1 isForeground=true target=Modal
1777871997770 consent: service.launchModal id=1 startActivity=ok
1777871997780 consent: trampoline.onCreate intent.id=1
1777871997780 consent: trampoline.bindIntent id=1 lookup=present
1777871997781 consent: trampoline.onResume id=1                      ← modal visible
1777872047315 consent: trampoline.unregisterModal id=1               ← user tapped Accept
1777872047331 consent: trampoline.onPause id=1 finishing=true
1777872057704 state ref=264523611 -> Cancelled(cause=PEER)            ← 60 s peer timeout
```

The modal opened, rendered, the user tapped Accept, the trampoline finished — but no `state -> Receiving` ever fired because the broadcast was dropped, so `submitUserConsent(true)` never reached the `InboundConnection`.

### Fix

`2f08a65` — drop `setClass`. `setPackage(packageName)` plus the unique action filter is enough to keep delivery scoped to our process; the dynamic receiver matches by action.

### Hardware verification (post-fix)

Same Galaxy S26 Ultra → Vivo X300 Ultra (V2547A on Funtouch 16) loop, this time after `7ee0816 + 2f08a65`:

```text
1777873108576 consent: trampoline.unregisterModal id=1
1777873108576 consent: trampoline.sendBroadcast id=1 accepted=true
1777873108592 sharing: sending RESPONSE ... state=ReceivingPayloads
1777873108594 state ref=43959953 -> Receiving(... totalSize=19 ...)
1777873108594 consent: dismiss id=1 prior=Modal
1777873108729 sharing: received PROGRESS_UPDATE ... bytes=15
1777873108730 state -> Receiving(bytesTransferred=19, totalSize=19, ...)
1777873108733 state -> Completed(items=[Text(payloadId=...,
                       data=[105,116,101,114,50,45,118,101,114,105,102,121,...])])
```

ASCII-decoding the `data=` byte array yields `"iter2-verify-143805"` — the exact text the Galaxy share intent submitted. **159 ms from Accept tap to `Completed`.**

### Consecutive sends

Triggering more sends back-to-back showed Galaxy's stock QS reporting `Failed` *Galaxy-side* (zero delta on Vivo's inbound log). The rapid-retry rate limit is on Galaxy's side and is independent of LibreDrop — `force-stop com.samsung.android.app.sharelive` between attempts is required for clean consecutive testing, and `pm clear` wipes the Private-sharing setup screen entirely. This matches what I noted in the original regression report.

### Pending follow-up (still LOW)

The two MEDIUM nits from the prior `pr-reviewer` pass (M1 surface-FSM rapid-flip race, M2 silent BAL fallback) remain documented as known limitations in the PR body. Neither is reproducible on the validated hardware path — they're guardrails for hostile OEMs and rapid foreground-flap races, not blockers for shipping.

PR is ready for re-review / re-merge.

kyujin-cho:
## Correction on the consecutive-sends note

Earlier I claimed Galaxy stock QS rate-limits retries to the same peer. That was a guess, not a finding — I should not have written it that way. Pulling Galaxy's actual logcat during the failed retry window (~14:39) shows a different and more interesting picture:

```text
NearbyMediums: Received advertisement header BleAdvertisementHeader {
    version=2, numSlots=2,
    serviceIdBloomFilter=[ 0x20 0x00 0x20 0x50 0x22 0x10 0x00 0x03 0x00 0x00 ],
    advertisementHash=[ 0xff 0x8a 0xa1 0x5c ],
    psm=0, supportExtendedAdvertisement=false }
NearbyMediums: Try to fetch advertisement from a possible IOS target.
NearbyMediums: Not allow Gatt actions during scanning.
```

Galaxy's scanner sees libredrop's `0xFEF3` advertisement on every BLE scan window and classifies it as `a possible IOS target`. While in that classification it refuses to initiate GATT actions. The GATT slot-read that drives the rest of the bootstrap therefore never starts on the failing retries.

I cannot yet conclude *why* this only manifests on the post-success retry, because the logcat buffer for the earlier successful send (~14:38) has rolled. To get a proper comparison I need a clean Galaxy QS state and a side-by-side capture, which I'll do in a follow-up debug-loop run.

**Scope of this PR is unchanged.** The iter-2 fix (`2f08a65`) is independently verified end-to-end via the successful send at 14:38: trampoline modal → Accept tap → `state -> Receiving` → `Completed(items=[Text(data=[105,116,101,114,50,...])])`. The "IOS target" mis-classification is a separate hypothesis for a separate bug; it does not affect the broadcast-routing fix this PR delivers.

Marking the consecutive-send result as "needs follow-up investigation" rather than "rate-limited."

kyujin-cho:
## Final correction on the "consecutive sends" finding

I was wrong twice before this; here is the actual root cause from `dumpsys wifi`:

```text
mWifiInfo SSID: <unknown ssid>, BSSID: <none>, IP: null,
  Supplicant state: DISCONNECTED
targetConfigKey="Lablup Office"WPA_PSK ... roam=false
  CMD_START_CONNECT  14:38:32   (around the working send at 14:38:09)
  CMD_START_CONNECT  14:39:57   (consecutive retry — dropped back to DISCONNECTED)
```

The Galaxy was unable to hold a stable Wi-Fi association during the test window. NearbySharing's medium availability check fails fast without a connected Wi-Fi:

```text
NearbyMediums: MEDIUM_ERROR [DEVICE][WIFI_LAN][START_DISCOVERING]
                            [MEDIUM_NOT_AVAILABLE][WITHOUT_CONNECTED_WIFI_NETWORK]
NearbyConnections: MediumManager failed to get IP address.
NearbyMediums: MEDIUM_ERROR [NETWORK][BLE][START_DISCOVERING]
                            [GATT_FETCH_ADVERTISEMENT_FAILED][TIMEOUT]
... but we have recently failed to read its GATT advertisement.
```

So:
- "Galaxy stock QS rate-limits" — wrong, no such mechanism here.
- "Galaxy classifies the ad as IOS target" — log noise, not the actual block.
- "Need Samsung Private Sharing setup" — wrong, that was just a screen `pm clear` brought up.
- **Actual cause**: Galaxy's Wi-Fi was DISCONNECTED. WIFI_LAN unavailable → discovery fails before BLE GATT bootstrap is even attempted. The first send happened to land in a brief associated window; every retry hit a DISCONNECTED window.

This is a test-bench flake, not a regression. The iter-2 fix in `2f08a65` is independently verified by the one successful end-to-end send — broadcast-routing was the only real bug in the foreground-modal path.

kyujin-cho:
## Consecutive-sends failure: not a libredrop bug

Pulled Galaxy's `dumpsys wifip2p` records and the full `NearbyConnections` logcat, and the cause turns out to be **Galaxy Sharelive policy**, not anything libredrop is doing wrong.

### The first successful send at 14:38

```text
P2P rec[5]  14:38:13.683 P2pDisabledState → InactiveState  what=WifiP2pManager.CONNECT  sender=com.google.android.gms
P2P rec[17] 14:38:14.184 GroupNegotiationState → GroupCreatedState  P2P_GROUP_STARTED_EVENT
```

Galaxy formed a Wi-Fi Direct group, the transfer completed, and at `14:41:16.321` the group was torn down (`DISABLE_P2P`).

### Every subsequent send since then

```text
P2P rec[94..99] 14:59:25 .. 15:00:33  REQUEST_GROUP_INFO  (com.google.android.mosey)
                                       — no CONNECT event, P2pDisabledState throughout
NearbyConnections (Galaxy): Init Wifi info start (allowTurnOnWifi:false).
NearbyConnections (Galaxy): client ... to endpoint XX4I disallows to turn on WiFi.
NearbyConnections (Galaxy): MediumManager failed to get IP address.
NearbyConnections (Galaxy): java.io.IOException: Failed to write ConnectionRequestFrame
NearbyConnections (Galaxy): Caused by: BleSocketOutputStream failed to write data
```

Galaxy's `BluetoothGatt` trace for the same window shows the BLE GATT bootstrap **succeeds**: connect, MTU=517, services discovered, indications enabled on the second-profile FROM_PERIPHERAL char, BLE EndpointChannel created. The `BleSocketOutputStream failed` exception is a **downstream symptom** — Galaxy's `MediumManager` gave up acquiring an IP for the eventual Wi-Fi transfer medium, closed the BLE stream, and the in-flight write failed because the stream had already been torn down.

### Why my earlier "this is #145 again" read was wrong

Issue #145's mechanism was "BLE socket written against the wrong `0xfef3` service handle." Here the BLE socket lands on libredrop's second-profile service correctly (`onSearchComplete status=0`, `setCharacteristicNotification 00000100-0004-1000-8000-001a11000102 enable=true` after `discoverServices` — exact same path the working 14:38 send took). The same exception text gets thrown for two completely different upstream causes; I conflated them.

### Root cause (this time with evidence)

Stock Sharelive on this Galaxy sets `allowTurnOnWifi:false` for the connection options after the first transfer in a session. Combined with Wi-Fi being radio-on but `Supplicant state: DISCONNECTED` and no AP, that gives:

- WIFI_LAN unavailable (no IP).
- WIFI_DIRECT requires a P2P CONNECT, which Sharelive won't initiate while `allowTurnOnWifi=false`.
- → no transport medium → connection aborted → BLE stream closed mid-write.

Workaround for further hardware testing on this device: associate Galaxy to a Wi-Fi AP, or toggle Wi-Fi off-then-on between sends. The `allowTurnOnWifi` policy resets when Wi-Fi state changes.

### Implications for PR #152

None. The iter-2 broadcast-routing fix in `2f08a65` is independently verified end-to-end by the one successful send at 14:38, where the modal opened, Accept tap fired the broadcast through `setPackage`-only routing, the inbound state reached `Receiving` 159 ms later, and `Completed(items=[Text(data=[105,116,101,114,50,...])])` landed with the correct payload. The post-first-send policy lockout on Galaxy is orthogonal.

kyujin-cho:
## Retraction: I do not know what controls `allowTurnOnWifi`

In the previous comment I claimed:

- *"Stock Sharelive on this Galaxy sets `allowTurnOnWifi:false` for the connection options after the first transfer in a session."*
- *"P2P bring-up requires this flag to be true."*
- *"Workaround: associate Galaxy to a Wi-Fi AP, or toggle Wi-Fi off-then-on between sends. The `allowTurnOnWifi` policy resets when Wi-Fi state changes."*
- *"This is a Galaxy / Sharelive policy decision, not a libredrop issue."*

None of those are backed by documentation. I searched the public [`ConnectionOptions.Builder`](https://developers.google.com/android/reference/com/google/android/gms/nearby/connection/ConnectionOptions.Builder) reference, the AOSP [nearby module](https://android.googlesource.com/platform/packages/modules/Connectivity/+/refs/heads/android13-qpr3-c-s10-release/nearby/), and several reverse-engineered Quick Share implementations. None of them describe an `allowTurnOnWifi` field, who sets it, what changes it, or whether Wi-Fi toggling resets it. The flag appears to live in proprietary GMS Nearby Connections code I do not have access to.

So all I can honestly say from this run is:

1. Galaxy's `BluetoothGatt` trace shows the BLE GATT bootstrap completes (`onClientConnectionState status=0 connected=true`, MTU=517, `onSearchComplete status=0`, `setCharacteristicNotification` on `00000100-0004-1000-8000-001a11000102 enable=true`, BLE EndpointChannel created).
2. Galaxy logs `Init Wifi info start (allowTurnOnWifi:false).` and `In requestConnection(), client X to endpoint XX4I disallows to turn on WiFi.`.
3. Galaxy logs `MediumManager failed to get IP address.`.
4. Galaxy throws `Failed to write ConnectionRequestFrame` / `BleSocketOutputStream failed to write data`.
5. Vivo's `libredrop-inbound.log` records zero new lines for the failed sends — so libredrop's GATT server never sees a write to its second-profile TO_PERIPHERAL char.
6. The successful send at 14:38 used WIFI_DIRECT as the upgraded medium (Vivo log: `medium-upgrade: server completed WIFI_DIRECT`).

What I cannot say without more evidence:
- Which component sets `allowTurnOnWifi:false`.
- Whether libredrop's endpoint metadata, advertisement, or BLE GATT response could influence whether Galaxy chooses to bring up Wi-Fi.
- Whether the failure is a Galaxy device-state issue, a libredrop interop issue, or a Galaxy + libredrop combination.

The iter-2 fix in `2f08a65` is verified by the single 14:38 successful end-to-end transfer (broadcast routing fixed, modal opens, accept lands `state -> Receiving -> Completed` with the correct payload bytes). Whether the consecutive-send failure is or is not in scope for this PR cannot be settled without running it down further; I should not have been confident either way.

kyujin-cho:
## Experiment result: RPA-rotation-from-mode-flip hypothesis falsified

Ran the controlled experiment from the previous comment: short-circuited `BleQuickShareAdvertiser.setAdvertiseMode` from `AppLifecycleScanModeObserver` so the advertiser is not torn down + re-registered on FG/BG transitions. Built, installed, ran 3 consecutive Galaxy → Vivo sends.

### Results

- **Send #1: success** end-to-end (modal → Accept → `Completed`).
- **Send #2: same `BleSocketOutputStream failed to write data` failure mode.**
- **Send #3: same.**

So the change was applied, but consecutive sends still fail. The simpler "stable advertise mode → stable RPA → consecutive sends work" hypothesis is wrong (or at least incomplete).

### What the experiment did confirm

Galaxy's `BluetoothGatt: connect()` target MAC was identical across send #1 and send #2 (`XX:XX:XX:XX:76:12`). So pinning the advertise mode did stabilize the address Galaxy connects to — that part of the mechanism in `BleQuickShareAdvertiser.kt:380-413` worked as expected. The failure is downstream of the address being stable.

### Where the two paths actually diverge

Both sends produced byte-for-byte identical pre-conditions up to and including `successfully created a BLE EndpointChannel`. The divergence is one log line later:

| Send | Post-EndpointChannel | Outcome |
|---|---|---|
| #1 | `wrote ConnectionRequestFrame` → UKEY2 → ... | `Completed` |
| #2 | **TID 11129 (NearbySharing orchestrator) emits `Closed BlockingQueueStream`** ~962 ms after `StartConnectToReceiver`. Same TID then logs `Waiting for weave socket onDisconnected callback to be called, timeout = 3000 ms.` `BleSocketOutputStream failed to write data` follows on the writer TID. | `Failed` (Galaxy UI) |

Between `StartConnectToReceiver` and `Closed BlockingQueueStream` on TID 11129 there are **no logged steps**. The decision to close happens inside proprietary GMS code with no externally observable trace. Without that source, I cannot say what condition flipped.

### Side observation (separate from the close decision)

`TID 11129` was concurrently attempting `BluetoothGatt connect()` to a **different** Vivo MAC (`XX:XX:XX:XX:D8:FE`) for slot-fetch, with `GATT_FETCH_ADVERTISEMENT_FAILED` `EXECUTION_EXCEPTION` `java.io.IOException`. So even with mode pinning, Galaxy is tracking ≥2 RPAs for libredrop's advertiser concurrently — one for slot-fetch (`D8:FE`, failing) and one for connect (`76:12`, succeeding). The remaining RPAs likely belong to libredrop's other simultaneous advertisements (legacy FEF3, extended FEF3, DCT FC73 — three independent `BluetoothLeAdvertiser` registrations, and AOSP's BLE stack draws a fresh RPA per registration).

That's a new (still-unconfirmed) hypothesis — **(H4)**: libredrop's three concurrent advertisements present three different RPAs to Galaxy, and Galaxy's per-RPA state-machine ends up in a stuck back-off after the first transfer. A fresh experiment to test (H4): disable two of the three advertisements and see if consecutive sends now work. I have not run this yet.

### State

Experiment reverted, working tree clean. Iter-2 fix in `2f08a65` remains independently verified by the single end-to-end successful send (in this run too — send #1 completed in this experiment). The consecutive-send failure is a separate, still-open investigation that probably wants its own issue rather than blocking PR #152's broadcast-routing fix.

kyujin-cho:
## Experiment H4 result: hypothesis untestable; investigation stalled

Ran two follow-on experiments to test (H4) "three concurrent advertisers each draw their own RPA, confusing Galaxy on retries":

### Experiment H4a — disable extended FEF3 + DCT, keep only legacy FEF3
Build clean, install on Vivo. Verified `dumpsys bluetooth_manager` shows ONLY libredrop's legacy `0xFEF3` (17-byte connectable). Result: **V2547A does not appear in Galaxy's picker at all on any of 3 send attempts.** Galaxy needs DCT for the visible-identity surface (per the docstring at `BleQuickShareAdvertiser.kt:67-73`).

### Experiment H4b — disable only the visible-extended FEF3
Re-enabled DCT, kept extended FEF3 disabled. `dumpsys` confirmed legacy `0xFEF3` (17-byte connectable) + DCT `0xFC73` (16-byte non-connectable) running, no 34-byte extended. Result: **V2547A still does not appear in Galaxy's picker.** Galaxy logcat shows it does receive the advertisement header (`advertisementHash=[ 0xd9 0x47 0xd0 0xc7 ]`) but immediately loops on `we have recently failed to read its GATT advertisement` — slot-fetch is failing without the extended advertisement up.

So removing the extended `0xFEF3` advertisement breaks slot-fetch entirely on this Galaxy/Vivo combo. I cannot test "does removing concurrent advertisers fix consecutive sends" because removing them breaks the first send.

### Why this is a dead end without different tools

The actual `Closed BlockingQueueStream` decision in NearbySharing's TID 11129 happens 962 ms after `StartConnectToReceiver` with **no logged steps in between**. That decision is in proprietary GMS code, and I have spent multiple iterations trying to isolate the cause from the outside without success. Each advertisement-shape change so far either:

1. Doesn't fix the consecutive-send failure (mode-pin experiment), or
2. Breaks first-send discovery entirely (advertisement-disable experiments).

Continuing to swap advertisement combinations is yielding diminishing returns — I am bisecting a problem whose decision boundary is invisible to me.

### Working tree

Reverted both experiments. `git diff --stat` is empty. PR remains at `2f08a65 fix(consent): drop setClass on modal-Accept broadcast` plus the diagnostic commit. Iter-2 fix is independently verified by the single end-to-end successful send under any of these advertisement configurations.

### Suggested next moves

- **Ship PR #152** on the strength of its proven fix; file a separate issue for the consecutive-send failure that captures everything we've learned (it is not a regression caused by this PR).
- For the consecutive-send issue specifically, the productive next steps need tools I don't have access to right now: (a) reverse-engineered `bjfs` / `dkog` / `dpbf` GMS code paths that own the close decision, (b) an instrumented build of GMS Nearby for visibility, or (c) a Pixel/non-Samsung Android device for cross-OEM comparison.

Marking the consecutive-sends investigation parked at this point.

kyujin-cho:
## Consecutive-sends investigation: root cause + workable fix identified

After many experiments, here is the experimentally-verified picture.

### Confirmed reproducible behavior

| Approach | Sends succeed |
|---|---|
| No intervention between sends | 1/N (only the first one) |
| `setAdvertiseMode` no-op (no FG/BG flips) | 1/N |
| Disable extended `0xFEF3` advertisement | 0/N (V2547A not even in picker) |
| Vivo Bluetooth toggle off/on | 1/N |
| Galaxy Bluetooth toggle off/on | **3/3 ✅** |
| Endpoint identity rotation + advertiser bounce (in-process) | 2/3 ⚠️ |
| Full ReceiverSession `stop+start` (in-process) | 1/5 |
| `Process.killProcess(myPid())` after each completion (relying on FG-service auto-restart) | **5/5 ✅** |

### What's actually wrong

The state contamination is on Galaxy's side, not libredrop's. Galaxy `Bluetooth toggle` clearing it confirms that — Vivo BT toggle and Vivo advertiser bounce don't.

Galaxy's `BluetoothGatt` trace on the failing send shows `connect → MTU=517 → (nothing for ~1.3s) → disconnect`. **No CCCD descriptor write, no characteristic write.** Galaxy connects, negotiates MTU, then silently aborts. The decision happens inside proprietary GMS Nearby code (TID 11129, `Closed BlockingQueueStream`) that emits no log between `StartConnectToReceiver` and the close.

Process restart on Vivo accidentally fixes the symptom because the fresh process generates new `randomMetadata()` in `EndpointInfo`, the GATT server registers under a new `app_if`, and the BLE advertiser draws fresh RPAs — collectively making libredrop look like a brand-new peer to Galaxy. Cumulative changes are necessary; partial in-process restarts (advertiser-only, identity-only, even full session-restart) leave just enough residual state that Galaxy still aborts.

### Workable fix-with-evidence

`android.os.Process.killProcess(android.os.Process.myPid())` after each `InboundConnectionCompletion`. With `START_STICKY` foreground service semantics on `connectedDevice` type, Android auto-restarts the service in ~1.5 s. Tested 5/5 consecutive sends successful with this approach.

Tradeoffs:
- ~1.5–2 s gap between sends where libredrop is briefly invisible.
- Foreground notification flickers across the kill.
- Doesn't survive on devices with aggressive OEM service-killers (vivo Funtouch's Security Care prompted us during install but didn't block auto-restart in this run; future Funtouch updates may add gating).
- Heavy-handed for what is fundamentally a Galaxy-side cache issue.

### What I cannot determine without different tools

The exact GMS-side condition that causes TID 11129 to abort. Reverse-engineered or instrumented GMS Nearby would be required. The fact that Galaxy Bluetooth toggle reliably clears the state confirms Galaxy keeps a per-peer cache that gets contaminated, but the cache key (endpoint id? RPA? GATT server handle? all of these?) cannot be confirmed from outside.

### Working tree

All experimental code reverted. `git diff --stat` empty. No commits proposed yet — this finding is information for deciding the next step:

1. Ship `kill-on-completion` as a workaround (~30 lines + a test). It works but is crude.
2. File the consecutive-sends issue as a separate ticket and ship PR #152 on its already-verified iter-2 fix.
3. Acquire reverse-engineered insight into GMS Nearby's per-peer cache logic and engineer a clean fix.

PR #169, opened 2026-05-05
feat: adapt outbound file chunk size

### LINKED ISSUE #168: enhancement: Make outbound FILE chunk size adaptive to payload size
## Problem / Background
`PayloadTransferEncoder.encodeFilePayload(...)` already accepts a `chunkSize`, but `OutboundConnectionDriver.streamOneFile()` currently hardcodes `PayloadTransferEncoder.DEFAULT_FILE_CHUNK_SIZE` (`512 KiB`) for every outbound FILE transfer.

That means all sends use the same frame size regardless of total payload size or the active transport. In a recent live `vivo -> Galaxy` transfer, LibreDrop upgraded successfully from BLE bootstrap to Wi-Fi Direct and completed the payload path, but throughput still plateaued around `~10 MB/s`. That session also showed the Galaxy hosting the Wi-Fi Direct group on `2467 MHz` (2.4 GHz), so radio/band choice remains the primary ceiling for that run. Chunk-size tuning is still worth pursuing, but it should be treated as a second-order throughput optimization rather than a fix for 2.4 GHz WFD sessions.

The protocol stack is already compatible with variable FILE chunk sizes:
- `core-protocol/src/main/kotlin/dev/bluehouse/libredrop/protocol/payload/PayloadTransferEncoder.kt` exposes `encodeFilePayload(..., chunkSize = ...)`
- `core-protocol/src/main/kotlin/dev/bluehouse/libredrop/protocol/payload/PayloadAssembler.kt` validates offsets, coverage, and the dedicated empty `LAST_CHUNK` terminator instead of assuming a fixed chunk size

Any new policy must keep enough headroom below `FramedConnection.SANE_FRAME_LENGTH` (`5 MiB`) for protobuf and crypto overhead.

## Proposed Solution
Replace the hardcoded sender chunk size in `OutboundConnectionDriver.streamOneFile()` with a deterministic selector that chooses a chunk size from the file's total size.

A reasonable first pass is a conservative tiered policy such as `256 KiB / 512 KiB / 1 MiB / 2 MiB`, with a hard cap that stays comfortably below the framed transport limit. The initial implementation can be based on file size alone; transport-aware tuning can remain a follow-up if needed.

Keep the existing dedicated zero-byte `LAST_CHUNK` terminator behavior unchanged, and add lightweight diagnostics so transfer logs show which chunk size was selected for each file.

## Acceptance Criteria
- [ ] `OutboundConnectionDriver.streamOneFile()` no longer hardcodes `PayloadTransferEncoder.DEFAULT_FILE_CHUNK_SIZE` for every FILE transfer
- [ ] A deterministic selector chooses FILE chunk size from total payload size and enforces a safe maximum below `FramedConnection.SANE_FRAME_LENGTH`
- [ ] The encoder still emits the dedicated empty `LAST_CHUNK` terminator, regardless of chosen chunk size
- [ ] Pure JVM tests cover the selector policy and at least one non-default chunk size path through the FILE encoder/assembler flow
- [ ] Manual large-file transfer verification records the selected chunk size and observed behavior over Wi-Fi Direct

## Technical Considerations
- Current fixed-size sender path: `core-protocol/src/main/kotlin/dev/bluehouse/libredrop/protocol/connection/OutboundConnectionDriver.kt`
- Current encoder default and parameterized API: `core-protocol/src/main/kotlin/dev/bluehouse/libredrop/protocol/payload/PayloadTransferEncoder.kt`
- Receiver-side compatibility and gap/coverage validation: `core-protocol/src/main/kotlin/dev/bluehouse/libredrop/protocol/payload/PayloadAssembler.kt`
- Transport frame safety cap: `core-protocol/src/main/kotlin/dev/bluehouse/libredrop/protocol/transport/FramedConnection.kt`
- Existing JVM tests already cover FILE payload framing and assembly behavior; extend those instead of introducing Android-only coverage for this policy

kyujin-cho:
## Summary
- replace the sender's fixed 512 KiB FILE chunk size with a deterministic file-size selector
- log the selected chunk size on each outbound file transfer and keep the tier cap well below the 5 MiB framed transport limit
- add JVM coverage for the selector policy, a selector-chosen non-default encoder/assembler round trip, and document where manual Wi-Fi Direct runs should record the chosen size

Closes #168

## Verification
- `./gradlew :core-protocol:test --stacktrace`
- `./gradlew staticAnalysis :discovery-android:testDebugUnitTest :app:testDebugUnitTest :app:assembleDebug --stacktrace`
- `git diff --check`

## Manual verification
- Device-side large-file Wi-Fi Direct verification was not run in this turn. The runbook now calls out the `LibreDropOutbound` `chunkSize=` log line to record during that pass.

### DIFF
--- a/core-protocol/src/main/kotlin/dev/bluehouse/libredrop/protocol/connection/OutboundConnectionDriver.kt
+++ b/core-protocol/src/main/kotlin/dev/bluehouse/libredrop/protocol/connection/OutboundConnectionDriver.kt
@@ -1225,9 +1225,9 @@ internal class OutboundConnectionDriver(
     }
 
     /**
-     * Stream a single [file] in 512 KiB chunks. Returns a non-null
-     * terminal result if the user cancelled mid-file; null on
-     * successful completion of this single file.
+     * Stream a single [file] in deterministic, size-based chunks.
+     * Returns a non-null terminal result if the user cancelled mid-file;
+     * null on successful completion of this single file.
      */
     private suspend fun streamOneFile(
         channel: SecureChannel,
@@ -1235,14 +1235,19 @@ internal class OutboundConnectionDriver(
     ): OutboundResult? {
         val source = file.openChannel()
         var chunkIdx = 0
+        val chunkSize = PayloadTransferEncoder.selectFileChunkSize(file.size)
         try {
+            logger(
+                "fsm: streamOneFile chunkSize=$chunkSize " +
+                    "name=${file.name} size=${file.size} payloadId=${file.payloadId}",
+            )
             val frames =
                 PayloadTransferEncoder.encodeFilePayload(
                     payloadId = file.payloadId,
                     fileName = file.name,
                     totalSize = file.size,
                     source = source,
-                    chunkSize = PayloadTransferEncoder.DEFAULT_FILE_CHUNK_SIZE,
+                    chunkSize = chunkSize,
                     lastModifiedTimestampMillis = file.lastModifiedTimestampMillis,
                     // `parent_folder` is also carried on every PayloadHeader,
                     // not just on FileMetadata. Quick Share receivers can
--- a/core-protocol/src/main/kotlin/dev/bluehouse/libredrop/protocol/connection/OutboundConnectionDriver.kt
+++ b/core-protocol/src/main/kotlin/dev/bluehouse/libredrop/protocol/connection/OutboundConnectionDriver.kt
@@ -115,6 +115,9 @@ internal class OutboundConnectionDriver(
     /** The most recent active payload, for progress UI. */
     private var currentItemPayloadId: Long? = null
 
+    /** Buffered inbound frames to service while the dispatch loop is streaming files. */
+    private var streamingWireChannel: Channel<OutboundWireMessage>? = null
+
     /**
      * Drive the entire sender-side lifecycle. Returns the terminal
      * [OutboundResult]; throws on coroutine cancellation (handled by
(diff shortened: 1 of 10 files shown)


kyujin-cho:
Validated the root cause on-device and updated the sender transport budget accordingly.

- Actual failure was not a generic large-file stall: upgraded `WIFI_DIRECT` transfers to stock Galaxy/GMS rejected oversized encrypted payload frames when the adaptive tier selected ~2 MiB chunks.
- Fix constrains the outgoing secure frame budget by transport medium and clamps upgraded `WIFI_DIRECT` file chunks back to the known-good 512 KiB path while preserving larger adaptive chunks for unconstrained media.
- Live validation on 2026-05-05 23:08-23:09 KST succeeded end to end from vivo `V2547A` to Galaxy `SM_S948N` for `Sonos_85.00.23.apk` (341064318 bytes). Sender log shows `medium-upgrade: client completed WIFI_DIRECT`, `chunkSize=524288`, `bytesSent=341064318`, and `safe-disconnect ... ack=true`; Galaxy log shows `TRANSFER_FINISHED` / `COMPLETE`.

Local verification:
- `./gradlew :core-protocol:test --tests 'dev.bluehouse.libredrop.protocol.payload.PayloadTransferEncoderTest' --tests 'dev.bluehouse.libredrop.protocol.connection.OutboundConnectionTest' --stacktrace`\n- `./gradlew :app:assembleDebug --stacktrace`\n- `git diff --check`

PR #170, opened 2026-05-06
feat: Korean localization and UI polish across send / receive / settings / credit

YEONFEEL96:
## Summary

Single visual language across every user-facing surface, plus first-class Korean support.

### Settings tab
- Redesigned into rounded card panels (save location, advertised name, background activity, shake-to-report) on a gray page surface.
- Status read-outs ("Current: …") rendered as chip pills instead of italic footnotes.
- Advertised-name `EditText` wrapped in a chip surface — replaces the bare-line underline.
- Borderless secondary buttons replaced with paired frosted pills (matches the QR panel's Close / Copy link layout).
- Bug-report toggle moved into the standard Material settings row (title-left + switch-right + supporting text below).
- Toolbar title swaps per tab — Home keeps the LibreDrop brand, Settings reads "Settings" / "설정".

### Credit page
- Same card-panel idiom (Developer / Polisher / QA / Special Thanks each in its own panel).
- Section headers stay in English on Korean devices ("Developer", "Polisher", "QA", "Special Thanks to") per product direction.
- Parami added to QA with avatar, bilingual tagline; wired into the staggered name-gradient sweep + smooth-corner avatar treatment.

### Consent trampoline (receiver)
- Type-aware completion title: "X is sharing images / videos / files with you" via mime classification of the announced payload.
- Consent file list center-aligned to match the centered "받을 파일" / "Files to receive" heading.
- Reject / Accept and Close / View image rows use `wrap_content` + parent `gravity=center` with a 140dp `minWidth` floor — keeps the group centered when locale label widths differ.
- Completed panel transition deferred until the MediaStore preview lookup resolves (600 ms timeout) — eliminates the brief "title + Close button only" frame that previously snapped to the preview state mid-render.
- Tapping View image now `finish()`es the trampoline so returning from the gallery drops the user beneath the dialog rather than back onto a stale completion popup.

### Send screen
- Device-name pill recentered against the screen's geometric axis (toolbar `contentInset` and asymmetric padding zeroed).
- Rejection bounce-back: when the receiver declines, the picker reappears under a transient toast-style banner ("X declined the file transfer") that fades out after ~3 s.
- "Can't find the device?" help link + Material `BottomSheetDialog` with two sections (visibility check + QR fallback). Replaces the inline same-Wi-Fi network hint card (now permanently hidden — the bottom sheet covers the same guidance with better discoverability).
- Status panel headline / target / message all center-aligned to share a single vertical axis with the photo card and PIN digits.
- Photo card elevation shadow no longer clipped — `clipChildren=false` + `clipToPadding=false` propagated through three ancestor containers.
- QR url field is now a chip with a `HorizontalScrollView` wrapper — full URL stays inside the viewport, swipe / long-press to copy. Same change applied to the standalone `ShowQrActivity` layout.

### Home tab
- "Always visible" pill rebrands to "받을 준비 완료" / "Ready to receive" with a state-aware caption below ("anyone nearby can share files with me" vs "scan-only" copy swapped by `SendReceiveFragment`).
- Send files / Send folder buttons share a 50/50 weighted row so their edges align with the visibility pill above on the same column axis.

### Korean localization
- ~140 string overrides under `values-ko/` plus a parallel `service-android/src/main/res/values-ko/strings.xml` for receiver foreground / consent / progress notifications.
- "Advertised Quick Share name" rephrased as "Quick Share 표시 이름" — the literal "광고" reads as a commercial advertisement to most Korean users; "표시" matches the Bluetooth-discovery intent.
- Subtitle no longer references "19 UTF-8 bytes" — replaced in both EN and KO with consumer-friendly phrasing about the name being shortened automatically for compatibility.

## Test plan

- [ ] Cold launch on Korean-locale device: every visible string reads in Korean; English locale falls back cleanly to the canonical copy.
- [ ] Send tab: device pill renders centered against screen midline regardless of label length; cancel / done buttons match the visibility pill's edges.
- [ ] Send tab → tap QR: URL chip clips inside the card, horizontal swipe scrolls the full URL, long-press → Select all → Copy yields the entire string.
- [ ] Send tab → "Can't find the device?": bottom sheet opens with two sections + Got it button; close button matches the picker's Cancel pill style.
- [ ] Send → receive declines: picker reappears with a transient banner above the Cancel button, then fades.
- [ ] Receive an image: completion popup appears in one transition (no "Close-only" flash), shows preview + Close + View image.
- [ ] Receive non-image: completion popup shows Close-only as before.
- [ ] Tap View image → gallery → press back: returns to home / launcher (the trampoline is dismissed).
- [ ] Settings tab: four card panels render with proper spacing; chip read-outs show current values; advertised-name input is a rounded chip.
- [ ] Credit screen: four card panels (Dev / Polisher / QA / Special Thanks); QA card lists Chiyak, TEQHNIKACROSS, Parami in order; section headers remain English on Korean device.


kyujin-cho (approved):
lgtm

PR #182, opened 2026-05-23
feat: add Japanese (ja) localization

YEONFEEL96:
## Summary

Adds Japanese (`ja`) string resources so the app renders in Japanese when the system locale resolves to Japanese (`ja`, `ja-rJP`); every other locale still falls back to the canonical English copy.

New files:
- `app/src/main/res/values-ja/strings.xml` — full UI: send/receive, settings, permission onboarding, share/sender flow, QR, consent + transfer-state copy.
- `service-android/src/main/res/values-ja/strings.xml` — receiver foreground, consent, and transfer-progress notifications.

## Approach

Mirrors the existing Korean (`values-ko`) override structure exactly:
- Same key set as `values-ko`, including the same **intentional non-translations** — proper nouns (`app_name`, contributor names, "Quick Share", "bada") and the `credit_section_*` wordmark headers render in English on Japanese devices too.
- All format placeholders (`%1$s`, `%1$d`, `%%`) and escaped glyphs (`\"`, `\n`) preserved verbatim.
- No build/manifest changes needed — there is no `resConfigs`/`localeConfig` restriction, so Android auto-selects the Japanese resources.

## Testing
- `./gradlew :app:assembleDebug` — builds, APK produced.
- `./gradlew staticAnalysis` — passes (ktlint + detekt).
- No `StringFormatInvalid`/`StringFormatCount` lint issues in the new translations.

### DIFF
--- a/app/src/main/res/values-ja/strings.xml
+++ b/app/src/main/res/values-ja/strings.xml
@@ -46,7 +46,7 @@
     <string name="credit_kevin_desc">マイルが足りません</string>
     <!-- YEONFEEL's tagline (Japanese rendering of the Korean source
          "아무도 알지 못하는 세상을 걷는 법"). -->
-    <string name="credit_yeonfeel_desc">誰も知らない世界の歩き方</string>
+    <string name="credit_yeonfeel_desc">だれもが識らない世界の歩き方</string>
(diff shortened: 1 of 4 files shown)


YEONFEEL96:

PR #183, opened 2026-05-23
feat: Quick Settings tile to toggle receiver discoverability

YEONFEEL96:
## Summary

Adds `BadaQuickShareTileService`, a Quick Settings tile that makes the device discoverable to nearby Quick Share senders straight from the system Quick Settings panel — no need to open the app.

Tapping the tile toggles the same process-wide `MdnsVisibilityOverrideHolder` "always visible" override that the in-app Send/Receive pill drives, and brings `ReceiverForegroundService` up/down to back it. ON publishes mDNS + the BLE fast advertisement unconditionally; OFF clears the override and stops the receiver.

## Details

- **State model:** the tile mirrors `MdnsVisibilityOverrideHolder.isActive`; `onStartListening` re-syncs each time the panel opens (standard listening tile).
- **Background FGS-start handling:** a Quick Settings tap is **not** on the platform's foreground-service start-from-background exemption list, so when the app is backgrounded `startForegroundService` can throw `ForegroundServiceStartNotAllowedException`. The tile catches it and falls back to opening `MainActivity`, which starts the receiver from a visible (exempt) context; the override set beforehand makes the gate advertise immediately.
- **Permission gate:** if the mandatory discovery permission is missing, the tile routes the user into the app (onboarding) instead of toggling a switch that can't advertise.
- **API 34:** uses the `startActivityAndCollapse(PendingIntent)` form on API 34+, the `Intent` form below.
- Manifest `<service>` with `BIND_QUICK_SETTINGS_TILE`, `QS_TILE` intent-filter, 24dp white vector icon, and `TOGGLEABLE_TILE` metadata. Strings added to `values` + `values-ko` (Japanese follow-up after the i18n branch merges).

## Testing
- `:app:assembleDebug`, `:app:ktlintCheck`, `:app:detekt` pass.
- Verified on a vivo device (Android 16): foreground direct-start path and background fallback path (`ForegroundServiceStartNotAllowedException` → open-app → receiver starts, gate publishes with `overrideOn=true`).

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -333,6 +333,34 @@
             android:theme="@style/Theme.Bada.ConsentDialog"
             android:label="@string/consent_activity_label" />
 
+        <!--
+          Quick Settings tile: toggles the receiver's "always visible"
+          override straight from the system Quick Settings panel so the
+          user can make this device discoverable to nearby Quick Share
+          senders without opening the app.
+
+          Required by the platform: exported=true plus the
+          BIND_QUICK_SETTINGS_TILE permission (only system_server, which
+          binds the tile, holds it) and the QS_TILE action filter. The
+          icon must be a 24dp solid-white VectorDrawable.
+
+          TOGGLEABLE_TILE marks this as a two-state on/off switch for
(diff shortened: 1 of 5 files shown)


YEONFEEL96:

PR #184, opened 2026-05-23
feat: prompt for full-screen-intent so background transfers pop the PIN

YEONFEEL96:
## Summary

Lets the incoming-transfer consent prompt (`ConsentTrampolineActivity`, with the PIN) pop full-screen over other apps / the lock screen when a transfer arrives while Bada is in the **background**.

The consent flow already attaches a full-screen intent to the consent notification, but on **Android 14+** `USE_FULL_SCREEN_INTENT` became a special access that is auto-denied for non-calendar/alarm apps. Without it the full-screen popup silently degrades to a heads-up notification. This PR adds the missing piece: detecting the grant state and routing the user to grant it.

## Details

- **`FullScreenIntentPermission`** — `canUseFullScreenIntent()` check (always true below API 34, where it's install-time) and a settings route (`ACTION_MANAGE_APP_USE_FULL_SCREEN_INTENT`, with notification-settings / app-details fallbacks).
- **First-launch dialog** in `MainActivity`, chained *after* the battery-optimization prompt so dialogs never stack; gated by a session latch + persisted dismissal flag (`FullScreenIntentPreferences`), and only when the access is actually requestable (API 34+ and not granted).
- **Settings-tab row** showing the current grant status with a button to the system page. Hidden entirely below API 34, where the permission is install-time.
- No changes to the consent notification itself — it already sets the full-screen intent, so granting the access is all that's needed for the popup to fire.

Strings added to `values` + `values-ko` (Japanese follow-up after the i18n branch merges).

## Testing
- `:app:assembleDebug`, `:app:ktlintCheck`, `:app:detekt` pass.
- Verified on a vivo device (Android 16): app builds/installs; the Settings row and first-launch prompt only surface when the access is ungranted on API 34+.

### DIFF
--- a/app/src/main/kotlin/dev/bluehouse/bada/MainActivity.kt
+++ b/app/src/main/kotlin/dev/bluehouse/bada/MainActivity.kt
@@ -26,6 +26,8 @@ import com.google.android.material.bottomnavigation.BottomNavigationView
 import dev.bluehouse.bada.battery.BatteryOptimizationOemHelper
 import dev.bluehouse.bada.battery.BatteryOptimizationPreferences
 import dev.bluehouse.bada.bugreport.BugReportFlowSupport
+import dev.bluehouse.bada.consent.FullScreenIntentPermission
+import dev.bluehouse.bada.consent.FullScreenIntentPreferences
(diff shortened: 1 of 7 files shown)


YEONFEEL96:

PR #185, opened 2026-05-23
feat: Quick Settings tile, background consent popup, and Japanese localization

YEONFEEL96:
Combined PR superseding #182, #183, and #184. They were split per feature, but the Japanese strings for the new tile / consent surfaces can only be added once their base keys exist, so bundling them keeps `values-ja` complete in a single landing. Each feature is a distinct commit below.

## 1. Japanese (ja) localization (was #182)
- New `values-ja` for `:app` and `:service-android` mirroring the Korean override structure (same keys, same intentional non-translations: proper nouns, `credit_section_*` wordmarks).

## 2. Quick Settings tile (was #183)
- `BadaQuickShareTileService` toggles the receiver's always-visible override (`MdnsVisibilityOverrideHolder`) and brings `ReceiverForegroundService` up/down, so the device can be made discoverable from the Quick Settings panel.
- A QS tap is not on the platform FGS-start-from-background exemption list, so the tile catches `ForegroundServiceStartNotAllowedException` and falls back to opening the app (a visible/exempt context). Permission-missing routes into onboarding. API 34 `startActivityAndCollapse(PendingIntent)` handled.

## 3. Background consent full-screen popup (was #184)
- `FullScreenIntentPermission` + a first-launch dialog (chained after the battery prompt) + a Settings-tab row, so on Android 14+ the user can grant `USE_FULL_SCREEN_INTENT`. Without it the incoming-transfer consent prompt (with the PIN) degrades to a heads-up notification instead of popping full-screen while Bada is backgrounded. Hidden below API 34 (install-time grant).

## 4. Japanese strings for the new surfaces
- `values-ja` entries for the `qs_tile_*` and `settings_fsi_*` / `fsi_dialog_*` keys added by features 2 and 3.

## Testing
- Verified on a vivo device (Android 16, ja-JP locale): tile foreground + background-fallback paths; credit screen renders the localized YEONFEEL tagline in Japanese; app UI renders in Japanese throughout.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -333,6 +333,34 @@
             android:theme="@style/Theme.Bada.ConsentDialog"
             android:label="@string/consent_activity_label" />
 
+        <!--
+          Quick Settings tile: toggles the receiver's "always visible"
+          override straight from the system Quick Settings panel so the
+          user can make this device discoverable to nearby Quick Share
+          senders without opening the app.
+
+          Required by the platform: exported=true plus the
+          BIND_QUICK_SETTINGS_TILE permission (only system_server, which
+          binds the tile, holds it) and the QS_TILE action filter. The
+          icon must be a 24dp solid-white VectorDrawable.
+
+          TOGGLEABLE_TILE marks this as a two-state on/off switch for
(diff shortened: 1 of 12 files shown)


kyujin-cho (approved):
lgtm

PR #186, opened 2026-05-24
fix: send-screen and credit-screen UI polish

YEONFEEL96:
A batch of small UI-polish fixes, each a self-contained commit. All localized in English / Korean / Japanese where text changed, and verified on a vivo device (Android 16).

## Send screen
- **File count / size split**: the payload header showed "N files • size" on one line, which collided with longer localized strings. Now the file count is the headline and the transfer size renders on its own smaller, secondary line below (hidden when size is unknown). Drops the unused `*_with_size` strings.
- **QR button restyle**: swapped the bold deep-blue gradient + white glyph for the app's frosted-glass surface (kept at the original 14dp rounded-square corners) with a secondary-gray tint, and replaced the glyph with the qr-code-2 icon.
- **Copy-link toast**: the "link copied" toast was suppressed on API 33+ assuming a system overlay; several OEM ROMs (vivo) don't show it, so it now always fires.
- **"Can't find the device?" sheet** — added a third section noting Bada is experimental: some peers may be discoverable but not transferable, support widens over updates.

## Credit screen
- **Title spacing**: zeroed `contentInsetStartWithNavigation` so the title sits naturally next to the back arrow instead of far from it.
- **Special Thanks note**: localized the previously Korean-only "(개발 토큰 제공)" note in all three locales (proper noun "Chiyak" untranslated).
- **Avatar border**: moved the 1dp identity border onto a MaterialCardView wrapper (outlined cards stroke evenly) holding a ShapeableImageView; both share the SmoothCornerTreatment so the border is uniform at the corners and photos still clip to the smooth-corner shape (the card alone fell back to a square content clip).

## Receiver
- **Blur backdrop fade-in**: the received-image blur backdrop snapped from GONE to VISIBLE once decoded; it now cross-fades in (alpha 0→1, 320ms) so it eases onto the dialog instead of popping.

## Testing
- On-device (vivo, Android 16): verified the send-screen split + QR button, copy-link toast, help-sheet section, credit title spacing, localized credit note, and even avatar borders (including full-bleed photo avatars).


kyujin-cho (approved):
lgtm

PR #187, opened 2026-05-24
feat: QR code / link sharing to stock Quick Share (#28, sender side)

YEONFEEL96:
Implements the Bada **sender** side of QR-code / link sharing so a file shown as a QR/link is actually delivered to a peer that opens it — verified end-to-end against stock Quick Share on a Galaxy Z Fold7.

## Background
The QR panel already generated a valid `https://quickshare.google/qrcode#key=...` URL, but nothing acted on it: the receive/match/connect path was deferred (#28). Opening the link on a Galaxy launches Google Quick Share, which derives an advertising token from the QR key and advertises it (hidden-mode TLV). This PR wires Bada to find that peer, prove QR ownership, and send.

## Changes
- **`QrHandshakeSigner` (core-protocol)** — signs the UKEY2 `authString` with the QR keypair's EC P-256 private key and returns the IEEE-P1363 (64-byte `R‖S`) signature used for `qr_code_handshake_data` (per PROTOCOL.md).
- **`OutboundConnection` / `OutboundConnectionDriver`** — take a `qrSigningKey: PrivateKey?` instead of static handshake bytes; the driver signs the authString at PairedKeyEncryption time and attaches the signature.
- **`SendActivity`** — persists the QR keypair for the panel session, derives the advertising/name keys, and on each discovery update matches resolved peers' EndpointInfo via `QrTlvMatcher`. On a match it auto-connects, threading the QR private key.
- **Wi-Fi LAN preference** — stock Quick Share advertises the QR token over BLE first, then upgrades the same endpoint to Wi-Fi LAN. Bada prefers a LAN-capable match (far more reliable than the BLE bootstrap), waiting a short grace period for the LAN route to surface before falling back to BLE.
- **`SendPeerPickerController`** — new `onPeersResolved` callback feeding the QR matcher.

## Verified on-device (vivo V2547A → Galaxy Z Fold7, both Android 16)
- Opening the Bada link launches Google Quick Share (`SamsungQrCodeActivity`).
- Bada derives the same keys and matches the Galaxy's hidden-mode token (`HiddenMatch`, recovering the device name).
- Connects over Wi-Fi LAN → UKEY2 → paired-key (the `qr_code_handshake_data` signature is accepted) → Introduction → file streamed → safe-disconnect acknowledged.
- The file lands in the Galaxy's Quick Share received folder.

## Testing
- `./gradlew staticAnalysis` and `:core-protocol:test` pass.

## Notes / follow-ups
- Receiver-side QR (Bada scanning someone else's QR) is still out of scope — this is the sender path only.
- A first attempt occasionally times out before the receiver accepts (transient); a retry succeeds. Hardening the BLE-only fallback and surfacing the recovered device name in the UI are good follow-ups.

### DIFF
--- a/app/src/main/kotlin/dev/bluehouse/bada/send/SendActivity.kt
+++ b/app/src/main/kotlin/dev/bluehouse/bada/send/SendActivity.kt
@@ -52,7 +52,11 @@ import dev.bluehouse.bada.protocol.connection.OutboundConnectionState
 import dev.bluehouse.bada.protocol.connection.OutboundResult
 import dev.bluehouse.bada.protocol.endpoint.DeviceType
 import dev.bluehouse.bada.protocol.endpoint.EndpointInfo
+import dev.bluehouse.bada.protocol.qr.DerivedQrKeys
+import dev.bluehouse.bada.protocol.qr.GeneratedQrKeyData
 import dev.bluehouse.bada.protocol.qr.QrKeyData
+import dev.bluehouse.bada.protocol.qr.QrKeyDerivation
+import dev.bluehouse.bada.protocol.qr.QrTlvMatcher
 import dev.bluehouse.bada.protocol.qr.QrUrl
 import dev.bluehouse.bada.service.receiver.AdvertisedDeviceNames
 import dev.bluehouse.bada.service.receiver.OutboundSessionActiveHolder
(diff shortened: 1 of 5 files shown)


kyujin-cho (approved):
lgtm

PR #188, opened 2026-05-25
feat(landscape): overhaul landscape UI for nav, send picker, and QR/PIN screens

YEONFEEL96:
## Summary

Landscape-mode UI overhaul across the main shell and the send flow so the
short landscape viewport stays usable without scrolling key controls off
screen. Portrait is unchanged except where noted (shared row/string tweaks).

## What changed

**Main shell / bottom nav (landscape)**
- Float the bottom navigation as a rounded **liquid-glass pill** in the
  bottom-left: tighter width (80% of natural), real backdrop blur on API 31+
  (software-capture + `RenderEffect`, translucent fallback below), elastic
  drag-follow selection capsule, tab icon+label centered in the capsule.
- Compact top bar via orientation-specific toolbar dimens.

**Send/Receive tab (landscape)**
- Split left (gallery preview) / right (controls) so everything fits without
  scrolling; compacted to also fit short landscape viewports.

**Send peer-picker (landscape)**
- Left half: title block + "Can't find the device?" help link; right half:
  scrollable peer list with the Cancel/Done button pinned bottom-right (16dp
  inset) as a frosted, outlined glass plate over the list.
- List clipped to the rounded card; rows fade into the gradient via fading
  edges; per-row press ripple spans the full card width with row text aligned
  to the header.
- Help bottom sheet: top-right "X" dismiss in landscape, fully expanded.

**QR / link panel (landscape)**
- Split into left QR code / right title + URL + buttons; QR sized for balance;
  disabled overscroll bounce.
- Dropped the "without approval" clause from the QR intro copy in en/ko/ja.

**PIN / status (landscape)**
- Status panel wrapped in a fill-viewport scroll so the awaiting-PIN block
  (phase + PIN + confirm message) centers when it fits and scrolls when it
  doesn't — text stays clear of the button and the top is never clipped.
  Slightly smaller PIN font for balance.

**Settings (landscape)**
- Extra bottom padding so the last item (shake-to-report) scrolls clear of
  the floating nav pill.

## Testing

- Manually verified on a non-GMS device (vivo / OriginOS) in landscape:
  nav pill, send/receive split, peer picker (incl. ripple/alignment), QR
  panel, help sheet, and a live awaiting-PIN state (text fully visible).


kyujin-cho (approved):
lgtm

PR #193, opened 2026-05-26
feat: add Traditional Chinese (zh-TW) locale

PeterDaveHello:
This pull request adds Traditional Chinese (Taiwan) translations for all user-visible strings in both the main Android app and its service module. This ensures that users with their device language set to zh-TW will see a fully localized experience, including all navigation, status, notification, and consent dialogs.


YEONFEEL96:
Thanks for your contribution.

Issue #195, opened 2026-05-28
Fix bugs


genus84:
**Unable to click "accept" or "reject" from the banner announcement, you have to go inside the application to click.**
Device: OnePlus 15, Android 16, ColorOS 16.0.7.207
<img width="1271" height="530" alt="Image" src="https://github.com/user-attachments/assets/26be3ad3-2eab-4761-bbca-508d3dd60645" />

YEONFEEL96:
Hello, we have confirmed that there are cases where the application is silently dropped after failing to pass the OS policy. We are currently working on fixing this implementation and fix this bug in next version if possible. Thank you for letting us know.

But because our main test device, vivo X300 Ultra makes full screen popup rather than banner-style notification, it's bit restricted to check if modification will resolve the issue. if issue repeats after the update, please let us know.

genus84:
Yeah just release the update soon, I will check

kyujin-cho:
Closing as https://github.com/kyujin-cho/Bada/releases/tag/v20260604.01 would solve this issue. Check it out and feel free to reopen this issue if the regression is not resolved.

Issue #190, opened 2026-05-25
Hopefully, Simplified Chinese can be supported.


Keepsmile31:
I am a Chinese user and my English is not very good. I hope you can support Chinese.

YEONFEEL96:
We will try to include the translation in the next version. Thanks.

pptext:
> 我们会在下一版本中尝试加入翻译。谢谢。

Thank you very much！！！

kyujin-cho:
#199 should resolve this issue. closing.

PR #199, opened 2026-05-31
fix: consent/cancel notification actions on Android 14+ and assorted copy/i18n cleanup

YEONFEEL96:
Bundled two different commit into one PR.
Resolves problem of #195 and #190 .

## 1. `fix(receiver): use setPackage-only routing for consent/cancel notification broadcasts`

The **Accept / Reject** buttons on the consent notification and the **Cancel** button on the in-progress transfer notification silently do nothing on stricter Android 14+ devices — reported on OnePlus 15 / Android 16 / ColorOS 16.0.7.207, and previously observed on Vivo Funtouch 16 (issue #151 and #195).

**Root cause.** Both notification builders construct their PendingIntent target with `Intent#setClass(context, ConsentBroadcastReceiver::class.java)`, scoping the broadcast to a specific receiver class. But `ConsentBroadcastReceiver` is registered **dynamically** at runtime by `ReceiverForegroundService` (with `RECEIVER_NOT_EXPORTED`) — there is **no manifest `<receiver>` entry**. Explicit-component lookup only resolves manifest-registered receivers; dynamic receivers are addressed via `IntentFilter` action match. The system finds no manifest target for the class name and silently drops the broadcast.

The same pitfall was previously found and fixed for `ConsentTrampolineActivity.sendBroadcast`. The notification PendingIntents kept the buggy form.

**Fix.** Drop the `setClass(...)` call in both notification builders. Keep `setPackage(packageName)` for in-process scoping; the dynamic receiver's `IntentFilter` matches `ACCEPT` / `REJECT` / `ACTION_CANCEL_TRANSFER`, and those actions are listened to only by `ConsentBroadcastReceiver` within the package, so action matching is unambiguous.

Files: `service-android/.../consent/ConsentNotification.kt`, `service-android/.../progress/TransferProgressNotification.kt`.

## 2. `chore(i18n): copy cleanup, empty-state padding, and zh-CN translations`

- **Settings buttons** (en/ko/ja): drop the trailing "…" on `main_save_location_pick`; rename `main_save_location_clear` and `main_advertised_name_reset` to a short "Reset" / "초기화" / "リセット" so the two buttons stop wrapping to a second line and breaking the settings card layout.
- **Empty-state copy** (en/ko/ja): trim the "Shared Wi-Fi is still the baseline path…" clause from `send_empty_state`.
- **Empty-state portrait padding regression**: the landscape send-picker refactor pulled the peer-list frame to full card width via negative side margins, which also stretched the empty-state TextView and the network-hint card to the card edges in portrait. Restore the 16dp horizontal inset so they align with the title again.
- **Simplified Chinese (zh-CN)**: add `values-zh-rCN/strings.xml` with 187 translated keys, mirroring the Korean / Japanese coverage. Brand names and tech acronyms kept in their canonical form; format placeholders and escapes preserved verbatim. (issue #190)

## Verification

- `./gradlew staticAnalysis` ✓
- `./gradlew :core-protocol:test :service-android:testDebugUnitTest` ✓
- `./gradlew :app:assembleDebug` ✓

### DIFF
--- a/app/src/main/res/layout/activity_send.xml
+++ b/app/src/main/res/layout/activity_send.xml
@@ -361,6 +361,7 @@
                                     android:layout_width="match_parent"
                                     android:layout_height="wrap_content"
                                     android:layout_marginTop="24dp"
+                                    android:paddingHorizontal="16dp"
                                     android:text="@string/send_empty_state"
                                     android:textAppearance="@style/TextAppearance.AppCompat.Body1"
                                     android:visibility="gone" />
@@ -376,6 +377,8 @@
                                     android:id="@+id/send_network_hint"
                                     android:layout_width="match_parent"
                                     android:layout_height="wrap_content"
+                                    android:layout_marginStart="16dp"
+                                    android:layout_marginEnd="16dp"
                                     android:layout_marginTop="16dp"
                                     android:background="?android:attr/colorBackgroundFloating"
                                     android:orientation="vertical"
--- a/app/src/main/res/values-ja/strings.xml
+++ b/app/src/main/res/values-ja/strings.xml
@@ -50,7 +50,7 @@
     <string name="credit_yeonfeel_desc">だれもが識らない世界の歩き方</string>
     <!-- Credit screen, QA section. -->
     <string name="credit_chiyak_qa_desc">コンサートのチケットが買いたいです</string>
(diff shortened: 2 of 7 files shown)


kyujin-cho (approved):
lgtm

PR #210, opened 2026-06-06
feat(tile): long-press opens app + larger Quick Settings tile glyph (#207)

### LINKED ISSUE #207: Feature Request: Long-press shortcut to open the main app interface
Currently, when I long-press the app icon in the quick settings/launcher, it redirects me to the system's "App Info" page.

I would like to request a feature where long-pressing the app icon instead opens the **main interface of the app**. This would provide a more convenient way to access the app directly.

Thank you for considering this improvement!

Image

YEONFEEL96:
Closes #207.

## Summary

Two small Quick Settings tile improvements requested in the issue, and for design polish:

1. **Long-press routes to the in-app surface.** Previously, long-pressing the Bada QS tile fell back to the system *App Info* page because no activity claimed `ACTION_QS_TILE_PREFERENCES`. Add an intent-filter for that action on `MainActivity` so long-press opens the app the user actually wants.
2. **Scale the tile glyph 1.2× from the viewport centre** so the icon reads larger inside the QS tile chrome. The 24×24 viewport is unchanged; the existing broadcast-mark path is wrapped in a `<group>` with `scaleX/scaleY=1.2, pivotX/Y=12`. Scaled paths still fit (outer ring tops out at ±9 from centre, 9×1.2 = 10.8).

## Verification

Verified on a real device (vivo PD2547 / Android 16):

- Short tap on the tile: still toggles always-visible (`MdnsVisibilityOverrideHolder`) and starts / stops the receiver — unchanged behaviour.
- Long-press on the tile: now opens `MainActivity` instead of the system App Info page.
- `adb shell cmd package resolve-activity --brief -a android.service.quicksettings.action.QS_TILE_PREFERENCES dev.bluehouse.bada` resolves to `dev.bluehouse.bada/.MainActivity`.
- Tile glyph is noticeably larger inside the QS chrome.

## Tests / Gate

- `./gradlew staticAnalysis` ✓
- `./gradlew :app:assembleDebug` ✓

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -243,6 +243,16 @@
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+            <!--
+              Long-press on Bada's Quick Settings tile (issue #207). The
+              platform fires ACTION_QS_TILE_PREFERENCES; without a matching
+              activity it falls back to the system App Info page. Routing
+              it to MainActivity opens the in-app surface the user actually
+              expects when they long-press the tile.
+            -->
+            <intent-filter>
+                <action android:name="android.service.quicksettings.action.QS_TILE_PREFERENCES" />
+            </intent-filter>
         </activity>
 
         <activity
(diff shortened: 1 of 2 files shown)


kyujin-cho (approved):
lgtm

PR #211, opened 2026-06-06
feat: check for updates against GitHub releases

YEONFEEL96:
## Summary

Adds an in-app "Check for updates" feature that queries the GitHub
Releases REST endpoint for the latest published Bada release and
surfaces it to the user without going through the Play Store.

## Why

There is no Play Store distribution channel for Bada, so users
have no way of knowing when a new APK is published. This PR closes
that gap: the launcher toolbar paints a red dot on the kebab the
moment a release newer than `BuildConfig.VERSION_NAME` shows up, and
a dedicated screen renders the new version + a one-tap jump to the
release page where the APK lives.

## What's new

**Update screen** (`update.CheckForUpdatesActivity`)

Reachable from the toolbar overflow menu next to **Credit**. Renders
four states off a single `StateFlow<UpdateState>` exposed by
`UpdateRepository`:

| State           | Title                         | Subtitle               | Primary button |
|-----------------|-------------------------------|------------------------|----------------|
| `Checking`      | "Checking for updates…"       | —                      | (disabled)     |
| `UpToDate`      | "You're on the latest version"| current version        | Check again    |
| `UpdateAvailable`| "New update available"       | latest version (small) | Update         |
| `Error`         | "Couldn't check for updates"  | error message          | Check again    |

The `Update` action opens the release URL via `ACTION_VIEW`.

**Cold-start refresh + dot indicator**

`MainActivity.onCreate` calls `UpdateRepository.seedFromCache()` (so
the dot is correct on the very first frame, even offline) and kicks
a one-shot `refresh()` that hits GitHub. The toolbar `overflowIcon`
swaps between `ic_overflow_kebab_24` and
`ic_overflow_kebab_with_dot_24`, and `onPrepareOptionsMenu` appends a
vertically-centred red dot to the menu item title via
`CenteredImageSpan` so the in-popup indicator matches the kebab dot.
Both follow the same StateFlow.

**Persistence**

`UpdatePreferences` caches the last-known release tag + URL in
SharedPreferences (project convention, not DataStore). Race between
`seedFromCache` and a concurrent `refresh` is guarded by a `Mutex`
inside the repository.

## Files

- `app/src/main/kotlin/dev/bluehouse/bada/update/`
  - `UpdateState.kt`, `UpdateRepository.kt`, `UpdateChecker.kt`,
    `UpdatePreferences.kt`, `CheckForUpdatesActivity.kt`,
    `CenteredImageSpan.kt`
- `res/drawable/ic_overflow_kebab_with_dot_24.xml`,
  `res/drawable/update_badge_dot.xml`
- `res/layout/activity_check_for_updates.xml`
- `MainActivity` wiring + overflow menu entry
- New translatable strings added to `values/`, `values-ko/`,
  `values-ja/`, and `values-zh-rTW/` per the add-to-every-locale
  policy

## Dependencies

None new. Uses `HttpURLConnection` + `org.json.JSONObject` (both
already in the Android SDK). `BuildConfig.VERSION_NAME` is used for
the installed-version side of the comparison; `buildConfig = true`
was added to `buildFeatures` to keep AGP 8 happy.

## Test gate

- `./gradlew staticAnalysis` — passes
- `./gradlew :app:testDebugUnitTest :app:assembleDebug` — passes
- Manual sanity checked on a vivo (non-GMS) device: cold-start
  badge dot, popup dot, screen state transitions, release URL jump.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -115,6 +115,10 @@ android {
 
     buildFeatures {
         viewBinding = true
+        // BuildConfig is opt-in on AGP 8+; UpdateRepository reads
+        // BuildConfig.VERSION_NAME to compare the installed app against
+        // the latest GitHub release tag, so it needs to be generated.
+        buildConfig = true
     }
 }
 
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -265,6 +265,18 @@
             android:parentActivityName=".MainActivity" />
 
         <!--
+          "Check for updates" screen reachable from MainActivity's
+          overflow menu. Runs a one-shot GitHub releases/latest query
+          via UpdateRepository / UpdateChecker and opens the matching
+          release page when the user taps "Update".
+        -->
+        <activity
+            android:name=".update.CheckForUpdatesActivity"
+            android:exported="false"
+            android:label="@string/update_activity_label"
+            android:parentActivityName=".MainActivity" />
+
+        <!--
           Sender share-intent target (#24).
 
           ACTION_SEND (single attachment) and ACTION_SEND_MULTIPLE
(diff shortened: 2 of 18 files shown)


kyujin-cho (approved):
lgtm

PR #233, opened 2026-06-15
Prettier incoming-transfer heads-up notification (bigger color-coded Accept/Decline + thumbnail)

IvanChanPing:
## What — prettier incoming-transfer heads-up notification
Redesigns the incoming-transfer consent pop-up so the **Decline** and **Accept** buttons are bigger, clearer, color-coded, and on opposite sides (Decline left, Accept right), with a thumbnail of what's coming in — so you can act on it at a glance instead of squinting at two small system buttons.

Prettier receive-consent heads-up notification

## How
- **`consent/ConsentNotification`** — builds the consent heads-up with a custom `RemoteViews` layout under `DecoratedCustomViewStyle` (Decline left / Accept right), a tinted small icon, and a thumbnail of the incoming item. Accept/Decline fire the **existing** `ConsentBroadcastReceiver` actions (`ACTION_ACCEPT` / `ACTION_REJECT`) and the body tap opens the **existing** `ConsentTrampolineActivity` via `ACTION_SHOW_CONSENT` — the consent flow itself is unchanged, only its presentation.
- **`consent/ConsentThumbnail`** — draws the placeholder thumbnail and the icon tints the layout uses.
- **`consent/ConsentNotificationStylePreferences`** — selects the notification style; **defaults to `RECOLORED`** (the redesign above). Two alternate layouts (`BRIDGE`, `SHEET`) sit behind this preference for experimentation, but the out-of-the-box default is the custom recolored view.
- **layouts** — `notification_consent` (the recolored view) + `notification_consent_bridge`.
- **drawables** — `notif_btn_accept_bg`, `notif_btn_decline_bg`, `notif_consent_bridge_card_bg`, `notif_consent_thumb_bg`.

This slots onto the existing receive-consent infrastructure already in the tree (`ConsentRegistry` / `ConsentRouter` / `ConsentCoordinator` / `ConsentNotificationContent` / `ConsentBroadcastReceiver` / `ConsentTrampolineActivity`) — the PR only adds the styled view and its resources, so it's self-contained and doesn't depend on the other Super Drop PRs.


YEONFEEL96:
My current test device won't let me to open this head-up notification, so it was unable to test it in real device.
Maybe I can tune a bit for the image icons and button edges to make it more refined.
Thanks for this improvement!

IvanChanPing:
> My current test device won't let me to open this head-up notification, so it was unable to test it in real device. Maybe I can tune a bit for the image icons and button edges to make it more refined. Thanks for this improvement!

That's surprising 

It's the same heads up notification that already existed just modified to be better.. what device did you test it on?

PR #243, opened 2026-06-28
feat: redesign the incoming-transfer consent heads-up notification

YEONFEEL96:
## Summary

Makes the incoming-transfer consent prompt show as a heads-up banner, and reworks its content so you can see what's incoming and confirm the PIN without expanding it.

### Heads-up vs full-screen

The full-screen intent and the ongoing flag are now only set when the screen is off or the device is locked; otherwise the prompt is a heads-up banner. On vivo (Funtouch/OriginOS) the full-screen intent was honoured regardless of lock state (the consent activity popped over whatever you were doing), and the ongoing flag suppressed the banner entirely (it only buzzed into the shade). The full-screen incoming-call sheet stays for the locked or screen-off case.

### Content

- The PIN sits next to the summary as its own element instead of trailing the summary text, so a long summary can't truncate the digits away.
- The summary breaks the file count down by type ("3 photos + 1 video") from the announced mime types. Stock Quick Share doesn't send preview thumbnails to non-GMS receivers (checked on-device: `preview_payload_ids` arrives empty even from a Galaxy sender), so this is the closest substitute.
- Expanding the notification shows a per-file list (one line per file with its size, capped with "...and N more"). The collapsed and heads-up views stay compact.
- The placeholder image is a line-art gallery vector tinted with `?attr/textColorSecondary`, so it stays visible on light and dark notification backgrounds.
- The action buttons are left-aligned and sized so the labels aren't clipped.

## Test plan

- Unit tests cover the content layer: PIN as a separate field, the photo/video breakdown, the file list and its "...and N more" line.
- `ktlintCheck`, `detekt`, the `service-android` tests and `assembleDebug` all pass.
- On a vivo device with a Galaxy Z Fold7 sending: it peeks as a heads-up banner when unlocked, the summary reads `사진 1장 (26.1 KB)`, and the gallery icon is visible on the card.


IvanChanPing:
I just wanted to note that center aligning the buttons was done because it was more in line with what iOS oppo Xiaomi etc did and might be easier for user.. (see screenshots). although left align might be more fitting with the theme of Google I personally think that having them centered and more in line with Apple and show me would make it just that much easier for the users to glance and click the right button. If you think left a line looked better I had an extensive notes to make it really easy to maneuver things

<br>
<br>

kyujin-cho:
Pushed review fixes directly to this branch (2518c3f, on top of a merge of current main):

- **PIN squeeze (high):** `notif_consent_body` now uses `0dp` + `layout_weight=1`, so a long summary (e.g. the folder-share path) ellipsizes instead of measuring the PIN caption/digits down to zero width.
- **Dismissible prompt:** `setOngoing(true)` is unconditional again — there is no deleteIntent/re-post path, so a stray swipe on an unlocked device would strand the sender until timeout. The keyguard/interactive gating of `setFullScreenIntent` (the actual vivo fix) is untouched.
- **Vector tint:** the gallery glyph's `?android:attr/textColorSecondary` tint is replaced with a fixed neutral `#FF8A8A8E` — theme-attr resolution inside RemoteViews is unreliable across OEM notification shades (#233 deliberately avoided it).
- **File list hardening:** blank-titled text items fall back to the kind name (same as the trampoline), and control characters in peer-supplied names are collapsed so an embedded newline can't forge extra list lines. `MAX_LISTED_ITEMS` is now 7 so the "…and N more" trailer survives one wrapped filename within the `maxLines=9` budget.
- **Cleanup:** removed the dead `consent_notification_body` / `consent_notification_bigtext_pin_line` strings from all locales, introduced a `consent_accent` color resource that deduplicates `#FF0A84FF` (layout, accept-button drawable, small-icon tint), and deleted the now-empty `ConsentThumbnail.kt`.

Stale "centered pair" comments were updated to describe the start-aligned button row, but the alignment itself is unchanged — that design question is intentionally left open for the ongoing discussion.

Validation: `staticAnalysis`, `:service-android:testDebugUnitTest`, and `:app:testDebugUnitTest` all pass locally (new tests cover the blank-title fallback and the newline-forgery case).

PR #246, opened 2026-06-30
Automatic update check + notification (download directly or open GitHub)

IvanChanPing:
## Summary
Extends the existing **Check for updates** feature (#211) with a proactive, automatic check. The app now polls GitHub Releases in the background every 6 hours and, when a newer version than the installed one is published, posts an **Update available** notification that lets the user open the release on GitHub or download and install it in place.

## What's new 
- A background check runs **every 6 hours** with no setup and no per-boot action.
- When a newer release exists, a notification appears: **"Bada update available — Version X is ready to install."**
- The notification's actions adapt to the release:
  - **View on GitHub** — always shown; opens the release page in a browser.
  - **Download & install** — shown **only when the release has an `.apk` asset attached**. It streams that APK into the system installer for a drop-in update. (When no APK is attached, only **View on GitHub** is offered.)
- Tapping the notification body opens the existing in-app Check-for-updates screen.

**In-app indicator:** the existing overflow-menu red dot is now also mirrored as a small red dot on the bottom-nav **Settings** tab, so a pending update is visible without opening the overflow menu. Both are driven by the same `UpdateRepository` state.

## How it works
- **`UpdateCheckWorker`** — a WorkManager `PeriodicWork` (`bada-update-check`, 6h, `NetworkType.CONNECTED`), scheduled in `BadaApplication.onCreate`. WorkManager persists the schedule across reboots, so the poll self-restarts on boot (no `BootReceiver` needed).
- **`UpdateChecker`** (existing) now also parses the release `assets[]` and exposes the first `.apk`'s `browser_download_url` (`LatestRelease.apkAssetUrl`), so the notification can decide whether a direct install is possible. The release endpoint is unchanged (`kyujin-cho/Bada` `releases/latest`).
- **`UpdateNotifier`** builds the adaptive notification (channel `app_update`).
- **`UpdateDownloadInstaller`** streams the APK into a `PackageInstaller` session (no temp file). The drop-in update works because the release ships the same package signed with the project key.
- **`UpdateInstallActivity`** (transparent trampoline) gates the one-time "install unknown apps" grant, then starts the install; **`UpdateInstallReceiver`** drives the system confirm dialog.
- The alert is de-duplicated per version (`UpdatePreferences.lastNotifiedVersion`), and an `autoCheckEnabled` flag (default on) is reserved for a future Settings toggle.

## Files
- `gradle/libs.versions.toml`, `app/build.gradle.kts` — add `androidx.work:work-runtime-ktx` 2.9.1.
- `app/.../update/UpdateChecker.kt`, `UpdatePreferences.kt` — extended.
- New: `UpdateNotifier.kt`, `UpdateDownloadInstaller.kt`, `UpdateInstallActivity.kt`, `UpdateInstallReceiver.kt`, `UpdateCheckWorker.kt`.
- `BadaApplication.kt` — schedule the worker.
- `AndroidManifest.xml` — add `REQUEST_INSTALL_PACKAGES`; register the trampoline activity + status receiver.
- `res/values/strings.xml` — `update_*` strings for the notification/install flow.

## Notes
- `POST_NOTIFICATIONS` (API 33+) is already declared/requested; if denied the notification is silently skipped (degraded, by design).
- Minimum WorkManager periodic interval is 15 min; 6 h is well within that and negligible for the GitHub unauthenticated quota.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,8 +106,8 @@ Networking notes:
 
 ## Permissions
 
-Bada asks only for permissions tied to receiving, discovery, and transfer
-visibility:
+Bada asks for permissions tied to receiving, discovery, and transfer
+visibility, plus one for its own update flow:
 
 - **Nearby Wi-Fi devices**: discovers and advertises Quick Share peers on the
   local network.
@@ -122,6 +122,13 @@ visibility:
   use physical location.
 - **Battery optimization exemption**: optional, but useful on OEM builds that
   aggressively stop background foreground services.
+- **Install unknown apps** (`REQUEST_INSTALL_PACKAGES`): lets the in-app
+  self-update install a newer Bada APK downloaded from GitHub releases; the
+  system still shows its own confirm dialog, and the grant is opt-in.
+
+Bada also polls GitHub releases every 6 hours in the background to notify you
+of new versions; this can be turned off with the **Automatic update check**
+toggle in Settings.
 
 ## Troubleshooting
 
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -169,6 +169,12 @@ dependencies {
     // ZXing (`zxing-android-embedded`) is intentionally not used.
     implementation(libs.zxing.core)
 
+    // WorkManager — runs the automatic update check (UpdateCheckWorker): a
+    // 6-hourly PeriodicWork, scheduled in BadaApplication.onCreate, that polls
+    // GitHub Releases and posts an "update available" notification. WorkManager
+    // persists the schedule across reboots with no user action.
+    implementation(libs.androidx.work.runtime.ktx)
+
     // NOTE: the self-ADB Wi-Fi stack (libadb-android + Conscrypt + BouncyCastle)
     // was MOVED OUT of :app into :radio-helper. The radios are toggled by the
     // standalone helper APK (which targets API 28 for the legacy capability and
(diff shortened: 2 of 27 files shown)


kyujin-cho:
Pushed review fixes directly to this branch (merge of `main` + `e66b789`). The maintainer decision is to **keep** the self-install capability; these commits harden it rather than remove it:

- **Download failures are now visible** — the failure notification is posted on its own id instead of being cancelled together with the progress notification.
- **No more leaked `PackageInstaller` sessions** — the staged session is abandoned on every failure path.
- **The download no longer runs on a bare `Thread`** — it moved into an expedited unique WorkManager worker (`UpdateInstallWorker`) with a foreground progress notification (dataSync FGS type on API 34+) and guaranteed cleanup on failure/cancellation.
- **Android 14+ BAL fallback** — if the install-confirm dialog can't be launched from the background receiver, a high-priority notification carries the confirm intent so the install can be finished from the shade.
- **Robust asset selection** — the release APK is matched by the `dev.bluehouse.bada-` filename prefix (releases also ship the radio-helper APK), falling back to "no update APK" when absent.
- **Settings opt-out added** — a new "Automatic update check" toggle in the Settings tab; disabling it cancels the persisted 6-hourly periodic work (not just a soft short-circuit).
- **Debug builds skip direct install** — `.debug` installs can never be updated by the release APK, so the action opens the GitHub release page instead.
- Deduplicated `isNewer` + notification-channel setup into shared helpers, translated the new strings to ko/ja/zh-rTW (zh-rCN intentionally left as-is — it carries no `update_*` strings at all yet), documented `REQUEST_INSTALL_PACKAGES` and the opt-out in the README, added manifest/unit tests, and removed the compile-only STATUS markers.

Validation: `./gradlew staticAnalysis :app:testDebugUnitTest :app:assembleDebug` is green (new tests: UpdateVersionsTest 4, UpdateApkAssetsTest 7, AndroidManifestPermissionsTest 16).

PR #247, opened 2026-07-01
Name Card: tap two phones to swap contacts

IvanChanPing:
## Name Card — tap two phones to swap contacts

Adds a **Name Card** feature: tap two phones back-to-back and, when neither is sending a file, they exchange contact cards — the same gesture as iPhone NameDrop.

### How it works
- **NFC is only the trigger.** A tap over a proprietary Name Card AID (`F0534443415244`) wakes both apps and hands over a 16-byte rendezvous token. It carries no contact data and is answered only while the device is unlocked (gated in code), so a locked phone never shares a card.
- **The card rides Bluetooth.** Using the token, the two phones connect over BLE GATT and exchange cards. The idle (tapped) phone runs the GATT server from a short foreground service; the phone with the app open is the reader/client.
- **Saving uses ContactsContract.** A received card is inserted directly when contacts permission is granted, otherwise it opens in the system Add-contact screen.

### What the user sees
- A **Name Card** row in Settings opens **My Name Card**, where you enter the name / phone / email to share (or fill it from your phone). A red dot on the row until a card is set up.
- On a tap, a full-screen card screen slides in over whatever was on screen: the received contact with **Receive Only** / **Share** (Share also sends your card back), or **Save** / **Done** on the tapped phone. The screen descends and expands in, and the AirDrop-style ripple plays on send.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -221,6 +221,20 @@
          killers from severing in-flight transfers. -->
     <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
 
+    <!-- Name Card (tap-to-share contacts): the "Use my phone info" button in the
+         My Name Card setup screen reads the device "Me" contact name and the SIM
+         line number to pre-fill the card. Both are runtime-requested on demand and
+         purely optional (the user can type the card manually); a denial is handled
+         gracefully. READ_PHONE_NUMBERS is the modern (API 33+) number read;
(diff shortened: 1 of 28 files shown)


IvanChanPing:
Don't implement yet I'm redesigning the whole thin

IvanChanPing:
https://github.com/user-attachments/assets/01d37027-b638-4f16-9cac-304c063e712b

kyujin-cho:
Closing this in favor of #251, which contains all of this branch's files plus the symmetric-consent redesign you mentioned — and its CI is green while this branch fails ktlint. Review feedback will continue on #251.

PR #251, opened 2026-07-03
previous version of namedrop rewritten

IvanChanPing:
previous version of namedrop rewritten

https://github.com/user-attachments/assets/01d37027-b638-4f16-9cac-304c063e712b


kyujin-cho:
Pushed a security-hardening commit (2ecfbe6) directly onto this branch (after merging current `main`). Summary of what changed and why:

**Token-authenticated consent is now the only protocol (C1/C2).** The v1 flow and the v2 "read-before-HELLO ⇒ legacy client, serve unconditionally" fallback meant any nearby BLE device could read the card with zero authentication and zero local consent. Now: the client must present the tap's 16-byte rendezvous token inside HELLO; the server verifies it in constant time (`MessageDigest.isEqual`) and disconnects any peer whose first consent message is not a valid token-bearing HELLO. The CARD characteristic answers `GATT_READ_NOT_PERMITTED` unless the peer is authenticated **and** the local user tapped Share; inbound card writes require the same authentication. A pre-token HELLO now decodes as malformed, so there is no downgrade path. `startServer`/`startClient` (v1) and the "Symmetric consent (beta)" toggle are removed — the old entry points (HCE-triggered service start, NDEF wake) all route into the authenticated flow.

**The token is never broadcast (M3-adjacent).** Advertising now carries only an 8-byte SHA-256 derivation of the token, so a BLE sniffer can't harvest the HELLO credential off the air — the token travels only over the proximity-bound NFC tap. This also let the service-data UUID move off the Bluetooth SIG member alias `0000fe2d-…` to a project-random 128-bit UUID (`918f0bd7-…`) while staying inside the 31-byte legacy advertisement (M3).

**Exported activity no longer trusts extras (C3).** `EXTRA_PEER_CARD` and `EXTRA_TOKEN` are gone. Peer cards and roles flow exclusively through the in-process `NameCardLinkHolder` session; the `NDEF_DISCOVERED` path (the reason the activity stays exported) only ever starts a fresh session from the OS-delivered NDEF token; any other launch finishes immediately. A hostile co-installed app can no longer inject a spoofed contact with a live Save button.

**Sharing is strictly opt-in.** `NameCardPreferences.isEnabled()` now defaults **false**, and `NameCardResolver` no longer falls back to the device "Me" contact / SIM line number — with no configured card, a tap exchanges nothing. The setup screen's "Use my phone info" button still pre-fills explicitly on user action (so `READ_CONTACTS`/`READ_PHONE_NUMBERS` stay). The `DISPATCH_NFC_MESSAGE` permission is removed: it is a signature-level platform permission not grantable to apps, and the "API 37 requires it" comment did not hold up against the SDK.

**Reader-mode scoping (H3).** `MainActivity` no longer arms NFC reader-mode on resume — that suppressed the app's own HCE (tap-to-open, Quick Share tap) and any wallet card whenever the home screen was open. The Name Card tap now rides the OS NDEF dispatch alone; `NameCardTapReader` is deleted with its only caller.

**Consent/commit correctness (H1/H2/H5).** The Share/Receive-Only commit latch flips and both buttons disable synchronously before any async work. The share-field filter now fails **closed**: if it can't produce a valid card, nothing is sent and the user sees an error (previously it silently sent the full card, deselected fields included). The server also now serves the field-filtered card instead of the full-card snapshot taken at startup. Config changes no longer break the flow: pending-save card, field selection, and the commit latch persist through rotation; the activity re-attaches to the live session instead of starting a second exchange; `restoreRadios(finishSession = isFinishing)` matches the send-screen pattern; `onNewIntent` is handled explicitly.

**Untrusted-input hygiene (M6/M7 + BLE races).** Peer cards are sanitized at the exchange boundary (reject >32 records, strip control/bidi-override codepoints, cap field lengths); diagnostics log only effect class names, never card fields. Scan results get a once-only connect guard, and `onCharacteristicRead` checks its status before parsing.

**Shader (M4).** The verbatim AGSL ripple with redacted provenance is replaced by original `ValueAnimator`/`ObjectAnimator` animations, so no unattributable third-party code ships under the project's Apache header.

**Validation.** `staticAnalysis`, `:app:testDebugUnitTest` (56 namecard tests incl. new token/sanitizer/codec suites), `:core-protocol:test`, and `:app:assembleDebug` all pass.

**Follow-ups (deliberately out of scope here):**
- Moving `NameCardConsentMachine`/`NameCardConsentCodec` (pure JVM) into `:core-protocol`.
- Translations (ja/ko/zh-rTW) for the Name Card strings, including the new fail-closed error string.
- **This feature remains device-unverified.** Per the repo's pre-merge policy, the two-device debug loop (vanilla-GMS + non-GMS) must run before merging — the BLE/NFC paths here are compile-only on the dev box.

PR #264, opened 2026-08-14
New Google-compatible Tap to Share content sharing from pixel 11

IvanChanPing:
## Summary

- add Google-compatible primary-AID Gesture Exchange content handoff (`A00000047609`) while preserving Name Card as a separate Bada-to-Bada feature
- normalize ordered `ACTION_SEND` / `ACTION_SEND_MULTIPLE` URI sources, preserve attachment precedence, and add UTF-8 text/link payloads
- stage unknown-length content off the main thread with size, free-space, stall, fsync, cleanup, and scavenging bounds
- carry files and text/BYTES through the existing Quick Share Introduction, transfer, consent, progress, completion, and teardown paths
- tighten one-role/one-tag ownership, radio-readiness arming, HCE prewarm, protocol/Noise validation, API-37 gating, reduced-motion rendering, and durable privacy-safe diagnostics

## Why this works with native Quick Share

This is not a Bada-to-Bada-only tap protocol. It implements Google’s native Gesture Exchange file-sharing path in both roles:

- when Bada sends, it acts as the NFC reader, selects Quick Share’s primary AID (`A00000047609`), completes the version, Noise-encrypted `SYSTEM_SHARE_SHEET`, and `nearby.sharing` capability handover, then connects using the Wi-Fi Direct credentials supplied by the native Quick Share peer
- when Bada receives, it exposes that same primary AID over HCE, returns its encrypted `nearby.sharing` identity and live Wi-Fi Direct credentials, and routes the resulting connection into Bada’s existing Quick Share receiver
- after the NFC bootstrap, files and text use the existing Quick Share Introduction, consent, payload, progress, completion, and teardown flow rather than a Bada-private transfer protocol

Therefore the intended peer is a stock Android phone with Google’s matching Tap to Share rollout enabled; the other phone does not need Bada installed. This is separate from Name Card, which remains Bada-to-Bada only. The wire implementation supports this interoperability path, while the physical two-phone acceptance test is still called out under Remaining verification.

## Dependency

This is a stacked draft based on the current head of #251. Review the final feature delta after #251 merges (or after this branch is rebased onto the resulting `main`). Until then, GitHub also displays #251’s unmerged history in this PR.

## Review notes

- Name Card remains SuperDrop-to-SuperDrop only; no Name Card/contact-card source is changed by this update.
- The older Quick Share NFC compatibility path remains the fallback when Google Tap to Share is disabled.
- API 37 types are isolated behind runtime SDK gates; minSdk remains 24. Polling annotations remain disabled unless physically validated.
- The diagnostic queue and private fallback are bounded, omit content/credentials/raw identifiers, and upload through a validated Internet network. Upstream can replace or disable the project collector endpoint before merge if repository policy requires it.
- The canonical debug APK and its SHA-256 are tracked in `CHANGELOG.md` for review.

## Validation

- `git diff --check`
- `./gradlew ktlintCheck detekt test assembleDebug` — successful across 332 Gradle tasks with no failing test XML
- intake tests cover URI-source ordering/deduplication, attachment precedence, and text/link routing
- protocol tests cover encrypted reader/responder flow, strict binding/metadata validation, corruption, replay/counter non-advance, plaintext limits, bidirectional traffic, and nonce exhaustion
- outbound loopback covers full text metadata and exact UTF-8 BYTES transfer into the existing inbound engine
- canonical APK is byte-identical to Gradle output, 21,501,621 bytes, v2-signed, SHA-256 `2523cb55b93a1cb82097a5a7da6c817c3bd506fa0330f1b40ef91c3de3c7e762`
- the exact APK was installed on Android 16; real UI gestures/clicks completed onboarding, opened Settings, displayed separate Name Card and Google Tap rows, toggled Google Tap off/on, rendered an HTTP link as a 32-byte “Sharing text” session, and cancelled back to Settings without an app crash marker

## Remaining verification

Physical interoperability is not claimed. The remaining acceptance gate requires two unlocked NFC phones, including a stock Pixel running the matching GMS rollout, to cover both directions, AID/RF negotiation, edge-light timing, recipient consent, file and text integrity, cancellation, retry, radio/helper failure, and cross-version fallback.

The broader design dossier also keeps these code gaps explicit: one actor does not yet own the complete content-attempt lifecycle; some capability semantics and external golden fixtures remain open; API-37 ChooserSession/one-transaction release are not implemented; and exhaustive lifecycle/accessibility/fault-injection coverage remains future work.


YEONFEEL96 (changes requested):
Thanks for this contribution — the Gesture Exchange work itself is substantial and clearly done with care (the Noise session with on-curve validation, the strict handover input validation, and keeping `core-protocol` free of `android.*` imports all check out). The interop direction is one we want. That said, there are several items that make the PR unmergeable in its current form. Listing them so they can be addressed one by one:

## Blockers

**1. Committed binary `bada-fork-debug.apk`**
The repo's `.gitignore` excludes `*.apk`; this file was force-added and is referenced only from the fork's `CHANGELOG.md`. A prebuilt binary is unreviewable and doesn't belong in the tree. Please remove it (the CHANGELOG SHA reference too) — a release/debug artifact can live on the fork's own Releases page if needed.

**2. Telemetry upload to a hardcoded IP endpoint (`GestureDiagnosticUpload`)**
Every tap event uploads a diagnostic line (plus a `X-Device: Build.MODEL` header) to `https://204-168-163-118.sslip.io/imelog/superdrop-tap-to-share` — an sslip.io hostname wrapping a raw IP with no attributable operator. It is on by default, has no consent prompt, ignores the existing bug-report opt-in, persists failed batches to disk and retries indefinitely, and deliberately selects any validated network including mobile data. For a privacy-positioned P2P app this cannot ship. Please remove the uploader entirely (local file logging like the existing `bada-inbound.log` pattern is fine), or rework it as strictly opt-in behind the existing diagnostics preference with a first-party endpoint the project controls.

**3. compileSdk/targetSdk bump to preview SDK 37**
`gradle/libs.versions.toml` moves compile/target to 37 and `gradle.properties` suppresses the unsupported-compileSdk warning, on AGP 8.7.3. That's a build-infrastructure change smuggled into a feature PR — CI (JDK 17) likely doesn't have the preview platform, and every other module inherits the risk. Please keep this PR on SDK 36 and isolate the API-37 NFC adapter behind runtime SDK gates only (which `Api37NfcSessionAdapter` already does); propose the SDK bump as its own PR when 37 is stable.

**4. Stale base — roughly half the diff is already on main**
The branch predates the squash-merge of #251, so all the NameCard code (~5,000 lines) reappears here as "new" and produces conflicts in `AndroidManifest.xml`, `SettingsFragment`, `fragment_settings.xml`, `strings.xml`, `MainActivity`, and `CHANGELOG.md` (which this PR replaces wholesale with fork-specific content). Please rebase onto current main so the diff shrinks to the genuine delta — for what it's worth, `ConsentTrampolineActivity`, `SendActivity`, and `themes.xml` apply cleanly on top of the recent consent-bottom-sheet rework, so the rebase should be mostly mechanical.

**5. Default-ON replaces the existing NFC paths**
`GestureTapToSharePreferences` defaults to `true`, which on upgrade silently (a) disables `BadaTapReader` on the send screen and (b) stops claiming the Quick Share AID (`F00000FE2C`) on the receive side in favor of the new HCE service — for every existing user, with no migration notice. Additionally, `maybeEnableTapReader()` now requires `shareRadios.isPrepared`, so on devices without the Radio Helper installed the tap reader may never arm at all — that looks like a regression independent of the toggle. Please default the switch to OFF until the interop path has passed the real-device matrix, and decouple NFC arming from radio-helper preparedness.

**6. Provenance of the GMS-derived artifacts**
`gesture_exchange.proto` is described as a wire-compatible subset of the Play services Gesture Exchange messages taken from "the selected GMS artifact", and `GestureEdgeGlowView` reproduces "the exact mapped GMS bezel light" geometry. Reimplementing a wire format for interop is normal for this project (that's what PROTOCOL.md is), but comments implying material was lifted from a decompiled GMS binary need to be reworded to describe observed wire behavior, and the glow visuals should be our own design rather than a copy of Google's.

## Required before merge (not architectural)

- All 15 new user-facing strings are English-only; the repo ships `values-ko`, `values-ja`, `values-zh-rCN`, `values-zh-rTW`. I can handle this part, so no need to handle the i18n right now.
- User-visible strings and the system-NFC service description still say "SuperDrop" (the fork's branding) — e.g. `settings_gesture_tap_subtitle`, `gesture_tap_hce_service_description`. Please rename to Bada.

Happy to re-review after a rebase with the blockers addressed. Items 1–3 are the hard ones; the feature itself we'd like to land.

IvanChanPing:
Correction (2026-09-19 18:16 UTC): The previous status wording was misleading. This comment records only the completed source and repository changes already pushed:

- The Google-compatible Tap to Share path is opt-in in Settings, so existing users retain Bada's legacy NFC path unless they enable it.
- NFC arming no longer waits for Radio Helper readiness; radio preparation is independent and best-effort.
- The committed APK and diagnostic uploader were removed, SDK 36 was restored, and Bada branding/provenance wording was corrected.

